PR #3, opened 2015-05-12
Adjust format for bearer token authorization

satnam6502:
No idea if this is right, but perhaps it will help me avoid getting this error:
```
2015-05-12 00:57:10 +0000 [info]: adding filter pattern="labels.**" type="kubernetes_metadata"
2015-05-12 00:57:10 +0000 [error]: config error file="/etc/td-agent/td-agent.conf" error="Invalid Kubernetes API 
endpoint: SSL_connect returned=1 errno=0 state=SSLv3 read server certificate B: certificate verify failed"
2015-05-12 00:57:10 +0000 [info]: process finished code=256
2015-05-12 00:57:10 +0000 [warn]: process died within 1 second. exit.
```


satnam6502:
Ah, I should have set `verify_ssl false` !

PR #4, opened 2015-08-24
Updating logic to retrieve cache_key

ewolinetz:
Previous logic wasn't pulling out values and resulting cache_key would be '___'  causing metadata values to be overwritten in the following block.

### DIFF
--- a/lib/fluent/plugin/filter_kubernetes_metadata.rb
+++ b/lib/fluent/plugin/filter_kubernetes_metadata.rb
@@ -123,7 +123,7 @@ module Fluent
         }
 
         if @kubernetes_url.present?
-          cache_key = "#{metadata['namespace']}_#{metadata['pod_name']}_#{metadata['pod_container_name']}"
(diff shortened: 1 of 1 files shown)


lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -123,7 +123,7 @@ def filter_stream(tag, es)
         }
 
         if @kubernetes_url.present?
-          cache_key = "#{metadata['namespace']}_#{metadata['pod_name']}_#{metadata['pod_container_name']}"
+          cache_key = "#{metadata[:kubernetes][:namespace]}_#{metadata[:kubernetes][:pod_name]}_#{metadata[:kubernetes][:pod_container_name]}"
```
jimmidyson: Should that just be `:container_name`?

ewolinetz:
@jimmidyson correct, that should have been just [:container-name].  Fixed.

jimmidyson:
Awesome! Thanks! This was lost in some refactoring & no test for correct caching...

PR #6, opened 2015-09-09
Add an include_namespace_id option that watches the namespaces in the

a-robinson:
cluster in order to be able to attach the namespace ID to log records if
requested.

Also change the field names from 'uid' to 'pod_id' and 'namespace' to
'namespace_name'. I'm happy to change either back if you'd like, but I
think this way is clearer.

I've verified this to work as desired in conjunction with my recent
out_google_cloud change.

I'll hopefully have time to add tests later tonight, but if not I'll get to them tomorrow morning.

@jimmidyson @mr-salty @brendandburns

### DIFF
--- a/lib/fluent/plugin/filter_kubernetes_metadata.rb
+++ b/lib/fluent/plugin/filter_kubernetes_metadata.rb
@@ -34,20 +34,20 @@ module Fluent
                  :default => '\.(?<pod_name>[^\._]+)_(?<namespace>[^_]+)_(?<container_name>.+)-(?<docker_id>[a-z0-9]{64})\.log$'
     config_param :bearer_token_file, :string, default: ''
     config_param :merge_json_log, :bool, default: true
+    config_param :include_namespace_id, :bool, default: false
 
-    def get_metadata(namespace, pod_name, container_name)
+    def get_metadata(namespace_name, pod_name, container_name)
       begin
-        metadata = @client.get_pod(pod_name, namespace)
-        if metadata
-          return {
-              uid:            metadata['metadata']['uid'],
-              namespace:      metadata['metadata']['namespace'],
-              pod_name:       metadata['metadata']['name'],
-              container_name: container_name,
-              labels:         metadata['metadata']['labels'].to_h,
-              host:           metadata['spec']['host']
-          }
-        end
+        metadata = @client.get_pod(pod_name, namespace_name)
+        return if !metadata
+        return {
+            namespace_name: namespace_name,
+            pod_id:         metadata['metadata']['uid'],
+            pod_name:       pod_name,
+            container_name: container_name,
+            labels:         metadata['metadata']['labels'].to_h,
+            host:           metadata['spec']['host']
+        }
       rescue KubeException
         nil
       end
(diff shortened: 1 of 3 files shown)


lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -67,7 +67,10 @@ def configure(conf)
       if @cache_ttl < 0
         @cache_ttl = :none
       end
-      @cache                                  = LruRedux::TTL::ThreadSafeCache.new(@cache_size, @cache_ttl)
+      @cache = LruRedux::TTL::ThreadSafeCache.new(@cache_size, @cache_ttl)
+      if @include_namespace_id
+        @namespace_cache = LruRedux::TTL::ThreadSafeCache.new([25, @cache_size].min, @cache_ttl)
```
jimmidyson: Limiting the cache size to 25 send a bit low. We have multitenant use cases with 100s of namespaces & setting it this low would cause lots of api server requests & cache evictions. I'd probably just make 25 the default but allow users to configure without hard ceiling.

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -34,20 +34,20 @@ class KubernetesMetadataFilter < Fluent::Filter
...
-              pod_name:       metadata['metadata']['name'],
-              container_name: container_name,
-              labels:         metadata['metadata']['labels'].to_h,
-              host:           metadata['spec']['host']
-          }
-        end
+        metadata = @client.get_pod(pod_name, namespace_name)
+        return if !metadata
+        return {
+            namespace_name: namespace_name,
```
jimmidyson: Why the move to `namespace_name`? Does id give us anything useful? Name is immutable & id isn't useful for a human POV.

jimmidyson:
@ewolitnetz Would this change affect anything you've done?

jimmidyson:
I'm not convinced that namespace id is that useful. Convince me ;)

a-robinson:
I'm not convinced that namespace id is useful either, given that the pod UID + container name already uniquely identifies the source of the logs across pod deletions and recreations. It's only really useful in the case that you delete and recreate your namespace if you want to be able to notice that everything in the new namespace afterwards is in a different namespace than before.

Unfortunately, I'm kind of stuck. I'm doing this is due to the schema requirements that the Google cloud logging/monitoring folks have put in place around using unique IDs rather than names. At this point, given that it's what monitoring is using, it's tough to justify changing it for logging. I'll revisit it with them, but more as feedback than as expecting a change.

If you don't want to take this addition, which is perfectly fair, my next step will be to move both the pod and namespace watching logic into the out_google_cloud plugin directly. That may be the best move anyway, given that it has the added benefit of cutting out all the json parsing necessitated by this being a separate plugin.

lib/fluent/plugin/filter_kubernetes_metadata.rb
a-robinson: I just got rid of the min setting -- even if there are enough namespaces around to fill the cache, it's nothing to worry about since entry each is just a pair of short strings.

lib/fluent/plugin/filter_kubernetes_metadata.rb
a-robinson: Regardless of the id issue, `namespace_name` closely parallels both `pod_name` and `container_name`.

As mentioned, though, I'm perfectly happy to name them whatever you'd like, as out_google_cloud can remap them without any issue, so just let me know what you'd like for both `namespace` and `uid`.

jimmidyson:
OK I'm happy it's behind guards so namespace ids will only be added if enabled.

Happy for metadata names to change too.

I'd like to keep this useful to the Google cloud logging fluent plugin as it means just a different storage mechanism to elasticsearch. Gives a bit of consistency across deployments.

Noticed that there's no test for the namespace id - could you add one please?

lib/fluent/plugin/filter_kubernetes_metadata.rb
a-robinson: Any preference here so I can update it when I add the test(s)?

jimmidyson: Let's go for namespace_name as you suggested.

a-robinson:
Thanks, @jimmidyson! Test added, along with a manually crafted VCR "recording".

jimmidyson:
Thanks @a-robinson - will get a release out.

PR #7, opened 2015-09-30
retain original log record from JSONL format

sosiouxme:
merge_json_log detects and parses JSON-format log lines in order
to add their fields as first-class citizens. However it was removing
the original log message and potentially overwriting other fields.
Now it ensures that the original log fields are retained.

### DIFF
--- a/lib/fluent/plugin/filter_kubernetes_metadata.rb
+++ b/lib/fluent/plugin/filter_kubernetes_metadata.rb
@@ -169,11 +169,7 @@ module Fluent
         log = record['log'].strip
         if log[0].eql?('{') && log[-1].eql?('}')
           begin
-            parsed_log = JSON.parse(log)
-            record     = record.merge(parsed_log)
-            unless parsed_log.has_key?('log')
-              record.delete('log')
-            end
+            record = JSON.parse(log).merge(record)
           rescue JSON::ParserError
           end
         end
--- a/test/plugin/test_filter_kubernetes_metadata.rb
+++ b/test/plugin/test_filter_kubernetes_metadata.rb
@@ -248,7 +248,7 @@ class KubernetesMetadataFilterTest < Test::Unit::TestCase
         'log' => "#{json_log.to_json}"
       }
       es = emit_with_tag('non-kubernetes', msg, '')
-      assert_equal(json_log, es.instance_variable_get(:@record_array)[0])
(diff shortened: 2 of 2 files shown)


sosiouxme:
@jimmidyson @ewolinetz

sosiouxme:
oh goodie, i broke tests.

ewolinetz:
+1

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -169,11 +169,7 @@ def merge_json_log(record)
         log = record['log'].strip
         if log[0].eql?('{') && log[-1].eql?('}')
           begin
-            parsed_log = JSON.parse(log)
-            record     = record.merge(parsed_log)
-            unless parsed_log.has_key?('log')
-              record.delete('log')
-            end
+            record = JSON.parse(log).merge(record)
```
jimmidyson: This flips the merge logic effectively from `record.merge(log)` to `log.merge(record)`. Which way round do we want to do this? Can't remember what the semantics for this are - only matters for conflicting keys I guess.

sosiouxme: merge overwrites duplicate keys. So here I think the right thing to do is
merge the original record over top of the parsed one. Besides keeping 'log'
it retains the stream and timestamp entries.

jimmidyson: Agreed.

therc:
Sorry to revive an old PR, but I'm not sure what the rationale for this is.
Suppose I output the line `{"foo": "bar"}`. Docker will write to disk something like `{"log": "{\"foo\":\"bar\"}\n", "stream":"stdout","time":"WHATEVER"}`. The plugin will rewrite that as `{"log": "{\"foo\":\"bar\"}\n", "stream":"stdout","time":"WHATEVER", "foo":"bar"}`. And once it's in ES, the entry will have

- `foo:"bar"`
- `log:"{\"foo\":\"bar\"}\n"`
- `_source:"FOO, LOG, ETC."`

Now there are three copies of the same data. That can get painful with large field values. One could drop `log` cluster-wide in a default mapping template, I guess. If that's a viable solution and the code is not going to change, should it at least be documented? It does have the drawback of also dropping lines that were never in JSON in the first place, e.g. stray debug print calls. For some that'll be a feature, for some a bug. :-)

PR #10, opened 2016-01-15
Use spec.nodeName for the host value in the metadata map

jchauncey:
fixes #9

PR #11, opened 2016-01-15
Use spec.nodeName rather than spec.host.

jimmidyson:


jimmidyson:
Supersedes #10

jchauncey:
can you let me know when you cut a new gem?

jimmidyson:
0.14.0 done already ;)

Issue #12, opened 2016-01-21
Use Kubernetes service account and KUBERNETES_SERVICE_{HOST,PORT} for automated configuration


joshk0:
First of all, good plugin, especially with the JSON parsing -- simple but effective.

Just an idea. Couldn't this plugin, if running within a Kubernetes pod, use the pod service account and host and port environment variables that are automatically provided to every pod to auto-configure API access?

jimmidyson:
That's a good idea. Try those env vars if available, enable enhanced metadata if can successfully query API server?

Would definitely accept a PR for this if you have time.

jimmidyson:
This is done now - thanks!

PR #13, opened 2016-01-22
Use Kubernetes services account if possible

joshk0:
If fluentd is running in a pod, it has sufficient access to the
Kubernetes API to enrich log records with pod and namespace metadata.
When URL / token / etc. are not implicitly specified, try and grab the
service account's credentials from the well-known directory.

Issue #12

### DIFF
--- a/fluent-plugin-kubernetes_metadata_filter.gemspec
+++ b/fluent-plugin-kubernetes_metadata_filter.gemspec
@@ -20,7 +20,7 @@ Gem::Specification.new do |gem|
 
   gem.required_ruby_version = '>= 2.0.0'
 
-  gem.add_runtime_dependency "fluentd"
+  gem.add_runtime_dependency "fluentd", "~> 0.12.0"
   gem.add_runtime_dependency "lru_redux"
   gem.add_runtime_dependency "kubeclient", "~> 0.4.0"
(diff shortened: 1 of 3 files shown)


stepanstipl:
+1 great, this would exactly solve my problem

jimmidyson:
LGTM. Can we add a test for this too? Similar to the test with the kubernetes URL set.

jimmidyson:
Oh & thanks very much!

joshk0:
Not sure what the best testing strategy was here since it involves environmental factors. Did the most reasonable thing i could think of, but you could argue that the ENV handling should be factored into an instance method instead... on the fence and don't often write Ruby, so not sure what's a best practice. Have a look.

joshk0:
Ping!

jimmidyson:
LGTM thanks!

jimmidyson:
FYI... released 0.15.0 to rubygems.org

joshk0:
Thanks!

jimmidyson:
Just a quick update @joshk0 - I found some minor issues with the logic for autoconfiguration, cleaned up & released 0.16.0. Worth upgrading.

joshk0:
Thanks for the heads up!

jimmidyson:
No worries - there's some other good fixes in there too, around label caching & proper support for dotted labels in ES 2.x.

Issue #16, opened 2016-03-03
Error for being unable to connect to k8s is convoluted


ewolinetz:
If the plugin is unable to connect to K8S, it doesn't print an error that is easy for users to interpret.

```
2016-03-03  03:32:39 +0000 [error]: unexpected error  error_class=Errno::ECONNREFUSED error=#<Errno::ECONNREFUSED:  Connection refused - connect(2)>
  2016-03-03 03:32:39 +0000 [error]: /usr/share/ruby/net/http.rb:878:in `initialize'
  2016-03-03 03:32:39 +0000 [error]: /usr/share/ruby/net/http.rb:878:in `open'
  2016-03-03 03:32:39 +0000 [error]: /usr/share/ruby/net/http.rb:878:in `block in connect'
  2016-03-03 03:32:39 +0000 [error]: /usr/share/ruby/timeout.rb:52:in `timeout'
  2016-03-03 03:32:39 +0000 [error]: /usr/share/ruby/net/http.rb:877:in `connect'
  2016-03-03 03:32:39 +0000 [error]: /usr/share/ruby/net/http.rb:1445:in `begin_transport'
  2016-03-03 03:32:39 +0000 [error]: /usr/share/ruby/net/http.rb:1402:in `transport_request'
  2016-03-03 03:32:39 +0000 [error]: /usr/share/ruby/net/http.rb:1376:in `request'
  2016-03-03 03:32:39 +0000 [error]: /opt/app-root/src/gems/kubeclient-0.4.0/lib/kubeclient/watch_stream.rb:20:in `each'
   2016-03-03 03:32:39 +0000 [error]:  /opt/app-root/src/gems/fluent-plugin-kubernetes_metadata_filter-0.16.0/lib/fluent/plugin/filter_kubernetes_metadata.rb:230:in  `start_watch'
   2016-03-03 03:32:39 +0000 [error]:  /opt/app-root/src/gems/fluent-plugin-kubernetes_metadata_filter-0.16.0/lib/fluent/plugin/filter_kubernetes_metadata.rb:140:in  `block in configure'
2016-03-03 03:32:47 +0000 [error]: fluentd main process died unexpectedly. restarting.
```

jimmidyson:
Yeah, not nice. Happy for a PR to improve that.

jimmidyson:
Closing - fixed in #17

PR #17, opened 2016-03-03
Rescuing on connection error within start_watch to be more informativ…

ewolinetz:
…e if K8S cannot be connected to

Addresses #16 

@jimmidyson PTAL


jimmidyson:
LGTM - thanks!

Issue #18, opened 2016-03-15
two "rake tests" are failing...both because the "labels" key/value pair has "labels" prefixed with ":"


stensonb:
```bash
/fluent-plugin-kubernetes_metadata_filter/test/plugin/test_filter_kubernetes_metadata.rb:307:in `block (2 levels) in <class:KubernetesMetadataFilterTest>'
<{:docker=>
  {:container_id=>
    "49095a2894da899d3b327c5fde1e056a81376cc9a8f8b09a195f2a92bceed459"},
 :kubernetes=>
  {:container_name=>"fabric8-console-container",
   :host=>"jimmi-redhat.localnet",
   :labels=>{"kubernetes_io/test"=>"somevalue"},
   :namespace_name=>"default",
   :pod_id=>"c76927af-f563-11e4-b32d-54ee7527188d",
   :pod_name=>"fabric8-console-controller-98rqc"}}> expected but was
<{:docker=>
  {:container_id=>
    "49095a2894da899d3b327c5fde1e056a81376cc9a8f8b09a195f2a92bceed459"},
 :kubernetes=>
  {:container_name=>"fabric8-console-container",
   :host=>"jimmi-redhat.localnet",
   :labels=>{:"kubernetes_io/test"=>"somevalue"},
   :namespace_name=>"default",
   :pod_id=>"c76927af-f563-11e4-b32d-54ee7527188d",
   :pod_name=>"fabric8-console-controller-98rqc"}}>

diff:
  {:docker=>
    {:container_id=>
      "49095a2894da899d3b327c5fde1e056a81376cc9a8f8b09a195f2a92bceed459"},
   :kubernetes=>
    {:container_name=>"fabric8-console-container",
     :host=>"jimmi-redhat.localnet",
?    :labels=>{:"kubernetes_io/test"=>"somevalue"},
     :namespace_name=>"default",
     :pod_id=>"c76927af-f563-11e4-b32d-54ee7527188d",
     :pod_name=>"fabric8-console-controller-98rqc"}}
```

the above output is from <code>rake test</code> and is showing a diff of the desired vs. expected output...the "labels" key/value pair is incorrect, as the "kubernetes_io/test" key is prefixed with a colon ":"

stensonb:
this is failing because the </code>.ruby-version</code> specifies "2.0.0":

```bash
[...]

PR #19, opened 2016-03-15
remove Gemfile.lock from gitignore to ensure other devs get the same …

stensonb:
…rubies when developing...

This might be why my <code>rake test</code> is failing in this ticket: #18 (this is now fixed by updating .ruby-version to "2.2.0" ; patch included in this PR)

Also, this PR just removes <code>Gemfile.lock</code> from gitignore, it does not add a <code>Gemfile.lock</code> -- maybe somebody can run <code>bundle install ; rake test<code> successfully, and submit another PR for a valid <code>Gemfile.lock</code>.  I would, but I'm unable to run rake test successfully...


jimmidyson:
As this is a rubygem it's normal to not commit the `Gemfile.lock` file. I'd rather follow that tbh.

As for the ruby version, do you think we should remove from this repo so devs can configure ruby how they want?

stensonb:
I suppose we'll just have to disagree on the <code>Gemfile.lock</code>...fwiw, bundler suggests committing: https://github.com/bundler/bundler/blob/1-0-stable/man/bundle-install.ronn#L248-L253

I would echo their comments by saying (anecdotally) I've run into more problems when using bundler without a Gemfile.lock than with...running "bundle install" on "tested" code without a Gemfile.lock is essentially throwing away all the previous test results (gem dependencies can change, your code can "stay the same", and the tests will, all of a sudden, break...no fun for anybody...especially a newcomer to the project.).

As for the .ruby-version...I'd say the only important part is that if a .ruby-version is specified, it should be used when running unit tests...and, that's not happening here (circle used 2.2.0)

jimmidyson:
I don't think that doesn't really matter/apply for rubygems, as this is, because it's installed more commonly via `gem install` on machines, which uses the gemspec to identify what dependent gems to install too. `bundle install` is used for development of the rubygem only & as such checking in can potentially hide issues with the gem dependencies range specified in the gemspec. I'd rather identify those earlier by contributors such as yourself identifying issues that come up in this way. I know this is a bit annoying but this is one of the problems with the distinction between `bundle install` & `gem install`.

jimmidyson:
Closing for the reasons discussed above.

Issue #22, opened 2016-04-19
Plugin should return string keys not symbols


richm:
The plugin returns this::

    {:docker:{:container_id:"46ac488dad3d826d86c1b095c8ea4b0904ee8c8c95e06e5f55571b99e91ee00f"},
     :kubernetes:{:namespace_name":"test",....}
    }

This is not permitted in fluentd: http://docs.fluentd.org/articles/plugin-development#record-format:
"Fluentd plugins assume the record is a JSON so the key should be the String, not Symbol. If you emit a symbol keyed record, it may cause a problem."

    router.emit(tag, time, {'foo' => 'bar'})  # OK!
    router.emit(tag, time, {:foo => 'bar'})   # NG!

This causes problems trying to pass the data through filters.  Specifically, record_transformer cannot reference the keys, either to use them to build other fields in the record, or to remove_keys.

richm:
closed - PR was merged

I've confirmed that the fix works with openshift/origin-aggregated-logging/fluentd

@jimmidyson Are you planning to do another release soon?

jimmidyson:
Thanks for confirming. Yeah I'll release tomorrow.

richm:
Thanks!

PR #23, opened 2016-04-19
Plugin should return string keys not symbols #22

richm:
https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/issues/22
The use of symbols for hash keys in the record is not allowed
in fluentd.  This commit adds a new plugin config parameter::

    use_string_keys false|true

The default is `false`, which means the metadata hash will have symbols
for keys.  If `true` is used, the hash keys will be strings instead
of symbols.

I recommend immediately deprecating the use of symbols as keys, and
make the default `use_string_keys true` as soon as possible.

For more information about why symbols should not be used, see
http://docs.fluentd.org/articles/plugin-development#record-format

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,8 +87,9 @@ Then output becomes as belows
 1. Fork it
 2. Create your feature branch (`git checkout -b my-new-feature`)
 3. Commit your changes (`git commit -am 'Add some feature'`)
-4. Push to the branch (`git push origin my-new-feature`)
-5. Create new Pull Request
+4. Test it (`GEM_HOME=vendor bundle install; GEM_HOME=vendor bundle exec rake test`)
+5. Push to the branch (`git push origin my-new-feature`)
+6. Create new Pull Request
 
 ## Copyright
   Copyright (c) 2015 jimmidyson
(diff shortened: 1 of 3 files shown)


richm:
@jimmidyson @sosiouxme @ewolinetz PTAL

We need this feature in order to be able to construct the hierarchical common metadata using a record_transformer filter.

jimmidyson:
Good catch. I'm happy to switch to string keys & document the change in the readme rather than adding in the config option & conversion. I can't imagine this will cause problems for any users.

ewolinetz:
LGTM

jimmidyson:
@ewolinetz Like I said I'd rather switch to string keys now that @richm has pointed out the fluentd requirement for string keys.

ewolinetz:
Sorry, i had missed that part of your comment... that sounds good to me

richm:
@jimmidyson @ewolinetz @sosiouxme Now the plugin only returns string hash keys - I removed the `use_string_keys` option

jimmidyson:
LGTM - thanks!

PR #24, opened 2016-05-28
support systemd-journald input format in addition to in_tail with wildcard

richm:
This will allow the plugin to look for CONTAINER_NAME in the journald input, and parse the k8s metadata from there.

### DIFF
--- a/lib/fluent/plugin/filter_kubernetes_metadata.rb
+++ b/lib/fluent/plugin/filter_kubernetes_metadata.rb
@@ -42,6 +42,14 @@ module Fluent
     config_param :secret_dir, :string, default: '/var/run/secrets/kubernetes.io/serviceaccount'
     config_param :de_dot, :bool, default: true
     config_param :de_dot_separator, :string, default: '_'
+    # if reading from the journal, the record will contain the following fields in the following
+    # format:
+    # CONTAINER_NAME=k8s_$containername.$containerhash_$podname_$namespacename_$poduuid_$rand32bitashex
+    # CONTAINER_FULL_ID=dockeridassha256hexvalue
+    config_param :use_journal, :bool, default: false
+    config_param :container_name_to_kubernetes_regexp,
+                 :string,
+                 :default => '^k8s_(?<container_name>[^\.]+)\.(?<container_hash>[a-z0-9]{8})_(?<pod_name>[^_]+)_(?<namespace>[^_]+)_(?<pod_id>[^_]+)_(?<pod_randhex>[a-z0-9]{8})$'
 
     def syms_to_strs(hsh)
       newhsh = {}
@@ -102,6 +110,7 @@ module Fluent
         @namespace_cache = LruRedux::TTL::ThreadSafeCache.new(@cache_size, @cache_ttl)
       end
       @tag_to_kubernetes_name_regexp_compiled = Regexp.compile(@tag_to_kubernetes_name_regexp)
+      @container_name_to_kubernetes_regexp_compiled = Regexp.compile(@container_name_to_kubernetes_regexp)
 
       # Use Kubernetes default service account if we're in a pod.
       if @kubernetes_url.nil?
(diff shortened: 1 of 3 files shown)


richm:
@sosiouxme @ewolinetz PTAL

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -42,6 +42,15 @@ class KubernetesMetadataFilter < Fluent::Filter
     config_param :secret_dir, :string, default: '/var/run/secrets/kubernetes.io/serviceaccount'
     config_param :de_dot, :bool, default: true
     config_param :de_dot_separator, :string, default: '_'
+    # if reading from the journal, the record will contain the following fields in the following
+    # format:
+    # CONTAINER_NAME=k8s_$containername.$containerhash_$podname_$namespacename_$poduuid_$rand32bitashex
+    # CONTAINER_FULL_ID=dockeridassha256hexvalue
+    config_param :journal_container_name, :string, default: 'CONTAINER_NAME'
```
t0ffel: Why do you need this? It is supposed to be 'CONTAINER_NAME' always, right?
on line 237 you use 'CONTAINER_NAME', not journal_container_name

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -163,55 +173,76 @@ def configure(conf)
...
-          if @include_namespace_id
-            namespace_name = metadata['kubernetes']['namespace_name']
-            namespace_id = @namespace_cache.getset(namespace_name) {
-              namespace = @client.get_namespace(namespace_name)
-              namespace['metadata']['uid'] if namespace
-            }
-            metadata['kubernetes']['namespace_id'] = namespace_id if namespace_id
-          end
-        end
+      metadata = tag.match(@tag_to_kubernetes_name_regexp_compiled) do |match_data|
```
t0ffel: My suggestion is to have a config parameter to decide whether the events are from journal or from log.
Then you can make the choice in `configure`, rather than when processing individual events.

richm:
@t0ffel @sosiouxme @ewolinetz  PTAL latest commit

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -220,13 +254,17 @@ def filter_stream(tag, es)
     end
 
     def merge_json_log(record)
-      if record.has_key?('log')
-        log = record['log'].strip
+      logkey = 'log'
```
jimmidyson: Can we set this in `configure` so we only need to do it once please.

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -163,55 +172,80 @@ def configure(conf)
...
-          }
-          if @include_namespace_id
-            namespace_name = metadata['kubernetes']['namespace_name']
-            namespace_id = @namespace_cache.getset(namespace_name) {
-              namespace = @client.get_namespace(namespace_name)
-              namespace['metadata']['uid'] if namespace
-            }
-            metadata['kubernetes']['namespace_id'] = namespace_id if namespace_id
-          end
+      unless @use_journal
```
jimmidyson: The logic for journald is pretty different to that for straight JSON docker logs because it doesn't use the fluentd tag for original container info so how about splitting this into separate functions for journald/json logging, determining which to use in `configure` based on the `use_journal` config option & then calling that in `filter_stream`? Gets away from the need for `unless @use_journal`/`if @use_journal`/etc which should give a bit of (probably unnecessary but nice to have) perf boost as well as being easier to reason about flows.

richm:
@jimmidyson does this latest commit address your concerns?

jimmidyson:
LGTM. If no more comments from @t0ffel @sosiouxme @ewolinetz then I'll merge tomorrow morning.

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -161,25 +170,34 @@ def configure(conf)
           end
         end
       end
+      if @use_journal
+        @merge_json_log_key = 'MESSAGE'
+        def self.filter_stream(tag, es)
```
jimmidyson: Actually I've just found out about [`alias_method`](http://ruby-doc.org/core-2.2.2/Module.html#method-i-alias_method) which seems a nice alternative to this dynamic def.

t0ffel:
LGTM

ewolinetz:
lgtm

jimmidyson:
Thanks @richm - do you need a release of this now I assume?

richm:
@jimmidyson re: new release - sure, whenever you're ready - Thanks!

jimmidyson:
0.22.0 [available](https://rubygems.org/gems/fluent-plugin-kubernetes_metadata_filter/versions/0.22.0) on rubygems for you now.

PR #25, opened 2016-06-02
not all container hashes are 8 chars long

richm:
also added some comments and removed some debugging code


lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -242,6 +242,7 @@ def filter_stream_from_journal(tag, es)
         metadata = nil
         if record.has_key?('CONTAINER_NAME') && record.has_key?('CONTAINER_ID_FULL')
           metadata = record['CONTAINER_NAME'].match(@container_name_to_kubernetes_regexp_compiled) do |match_data|
+            # container_hash, pod_id, and pod_randhex from the regex match_data are currently unused
```
jimmidyson: Perhaps these should be non capturing groups in the regexp then?

richm:
@jimmidyson I had to get rid of the named groups, because you can't have named groups that are non-capturing, but I documented what the fields are in the comments.

jimmidyson:
:+1:

Issue #28, opened 2016-06-14
plugin stops detecting changes in pods after some time


t0ffel:
I'm running openshift 3.2
the plugin is deployed as part of aggregated logging.

Successful case:
1. Deployed new fluentd pods with this plugin.
2. Deployed another test pod with specific set of labels.
3. Checked that test pod's labels are correctly reflected in kibana.
4. Edit test pod: add one label
5. Check that test pod's labels are correctly updated in kibana.

Failure:
1. Deployed new fluentd pods with this plugin.
2. Deployed another test pod with specific set of labels.
3. Checked that test pod's labels are correctly reflected in kibana.
4. Wait for 1-2 hours.
5. Edit test pod: add one label
6. Check that test pod's labels are correctly updated in kibana. (failure: labels are not updated).

What i think is happening: watcher times out and stops receiving any notices from k8s.
Please let me know what you think.

cc @richm

jimmidyson:
Very possible. Looks like the watch thread doesn't loop if watch closes. PR very welcome.

jcantrill:
@t0ffel have you seen and resolved a similiar issue in https://github.com/ViaQ/fluent-plugin-openshift_metadata_input  ?   Is it something you could patch in since its obviously an issue we are experiencing now?

vvk:
@jimmidyson, any thoughts about #29 which proposes the fix for this issue? This one is really critical for the scenarios that rely on a real-time update of annotations on namespaces.

jcantrill:
Closing stale though watch has gone through changes and may have resolved this issue.

PR #29, opened 2016-06-15
wrapped watcher in a loop, so that when k8s times out watcher starts anew

t0ffel:
I've just wrapped the watcher in the loop. this fixed the issue for me. is this what you had in mind?

fixes https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/issues/28


jcantrill:
@t0ffel Can you rebase and make a PR against openshift logging, to pull in this change for testing.  Possibly we can shorten the watch time or something to vet the change.

jcantrill:
closing in favor of #204

Issue #33, opened 2016-07-28
emit transaction failed: error_class=NoMethodError error="undefined method `strip\' for #<Symbol:0x000000033f370c>


olalonde:
I apologise in advance if I'm posting this to the wrong repo. I recently started using Kubernetes and keep seeing the following error in my logs:

``` javascript
{ log: '2016-07-28 14:47:55 +0000 [warn]: emit transaction failed: error_class=NoMethodError error="undefined method `strip\' for #<Symbol:0x000000033f370c>" tag="kubernetes.var.log.containers.blockai-ui-v6-web-7a4hq_blockai-ui_blockai-ui-web-39b0262bb5ba550fa6e7d824d337ff3b3cdb77d4c0a50b4221405cdae2be0475.log"\n',
  stream: 'stdout',
  docker: { container_id: '70edf98b8595202c6b564010fe66e8348168cdedeccbf0fd9a94decb0caf39d1' },
  kubernetes:
   { namespace_name: 'kube-system',
     pod_id: '4202b066-52eb-11e6-a5e2-068755380eff',
     pod_name: 'fluentd-elasticsearch-ip-172-20-0-130.us-west-1.compute.internal',
     container_name: 'fluentd-elasticsearch',
     labels: { 'k8s-app': 'fluentd-logging' },
     host: 'ip-172-20-0-130.us-west-1.compute.internal' } }
```

Is it possibly an issue with this plugin? Thanks!

Possibly related: https://github.com/kubernetes/kubernetes/issues/29640

jimmidyson:
Can you let me know what version of the plugin you're using? You can check either by looking at the dockerfile or by execing into the container & perhaps issuing gem list to see. Pretty sure we moved away from symbol keys a little while ago so very possibly just upgrading the plugin should do it.

olalonde:
Thanks, tried to inspect the container a bit but couldn't find any gems.

```
kubectl --namespace=kube-system describe po fluentd-elasticsearch-ip-172-20-0-131.us-west-1.compute.internal
Name:           fluentd-elasticsearch-ip-172-20-0-131.us-west-1.compute.internal
Namespace:      kube-system
Node:           ip-172-20-0-131.us-west-1.compute.internal/172.20.0.131
Start Time:     Mon, 25 Jul 2016 21:41:53 -0700
Labels:         k8s-app=fluentd-logging
Status:         Running
IP:             10.244.2.2
Controllers:    <none>
[...]

PR #34, opened 2016-08-04
Only call strip() if the object has that method.

edevil:
Symbols don't have it and sometimes we get those instead of strings.


edevil:
Fixes #33.

edevil:
The bug has been fixed upstream. https://github.com/fluent/fluentd/pull/1147

PR #36, opened 2016-09-06
Fix compatibility with fluentd v0.14

Malet:
v0.14 requires a filter method, it implodes if there isn't one


repeatedly:
Late reply. You should define `filter_stream` instead of `filter`.
Latest fluentd optimizes filter pipeline when all filter plugins use `filter` or `filter_with_time` method.
kubernetes_metadata_filter uses `filter_stream`, so you should use `filter_stream` for placeholder, not `filter.`

jimmidyson:
@repeatedly Aha! That makes a lot more sense. I was confused before... Thanks!

PR #37, opened 2016-09-08
Define filter_stream instead of filter

jimmidyson:


fluent-plugin-kubernetes_metadata_filter.gemspec
```diff
@@ -18,7 +18,7 @@ Gem::Specification.new do |gem|
   gem.require_paths = ["lib"]
   gem.has_rdoc      = false
 
-  gem.required_ruby_version = '>= 2.0.0'
+  gem.required_ruby_version = '>= 2.2.0'
```
repeatedly: Is this intended change?
This means this plugin doesn't work with current td-agent, google-fluentd or ruby 2.1 based system.

jimmidyson: Doh sorry! Will have to revert later & release as I'm out for a while now.

jimmidyson: 0.25.2 released with reverted requirement.

PR #44, opened 2016-10-17
Various fixes:

### LINKED ISSUE #43: Why do we cache pod metadata using the container name?
From https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/blob/097c60d5546011733bb5d6ba53160a89a3697122/lib/fluent/plugin/filter_kubernetes_metadata.rb#L222, we can see that the cache key used for the object being cached includes the container name.

Yet, `get_metadata` method that fetches the metadata object to be cached (https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/blob/097c60d5546011733bb5d6ba53160a89a3697122/lib/fluent/plugin/filter_kubernetes_metadata.rb#L74) does not use the contain name at all, though it does add it to the object returned.

This seems to imply that we'll have multiple versions of the same pod metadata cached under different keys.  In turn, this can lead to unnecessary traffic to the API server.

Changing the code to fetch the metadata blob using a cache key only using namespace name and pod name, will allow us to still add the container name to the object returned while reducing API server traffic, and reducing duplicate data cached.

Further, if the traffic drops significantly, one could imagine making the connections to the API server timeout to avoid the need for a continuous connection just for logs.

jimmidyson:
* Fixes #43: only use namespace name and pod name as cache keys (cc @portante)
* Delete empty matched annotations on update
* De-dot annotations as we do for labels

### DIFF
--- a/lib/fluent/plugin/filter_kubernetes_metadata.rb
+++ b/lib/fluent/plugin/filter_kubernetes_metadata.rb
@@ -79,6 +79,7 @@ module Fluent
         annotations = match_annotations(syms_to_strs(metadata['metadata']['annotations'].to_h))
         if @de_dot
           self.de_dot!(labels)
+          self.de_dot!(annotations)
         end
         kubernetes_metadata = {
             'namespace_name' => namespace_name,
@@ -372,6 +373,7 @@ module Fluent
               annotations = match_annotations(syms_to_strs(notice.object.metadata.annotations.to_h))
               if @de_dot
                 self.de_dot!(labels)
+                self.de_dot!(annotations)
               end
               cached['kubernetes']['labels'] = labels
               if annotations.empty?
--- a/lib/fluent/plugin/filter_kubernetes_metadata.rb
+++ b/lib/fluent/plugin/filter_kubernetes_metadata.rb
@@ -374,7 +374,11 @@ module Fluent
                 self.de_dot!(labels)
               end
               cached['kubernetes']['labels'] = labels
-              cached['kubernetes']['annotations'] = annotations unless annotations.empty?
+              if annotations.empty?
+                delete(cached['kubernetes']['annotations'])
+              else
+                cached['kubernetes']['annotations'] = annotations
+              end
               @cache[cache_key] = cached
             end
           when 'DELETED'
--- a/lib/fluent/plugin/filter_kubernetes_metadata.rb
+++ b/lib/fluent/plugin/filter_kubernetes_metadata.rb
@@ -71,7 +71,7 @@ module Fluent
       newhsh
     end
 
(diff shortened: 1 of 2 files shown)


jimmidyson:
Note changes to test are only for fixing of de-dotting annotations, change of cache key didn't require test changes so functionality unaffected.

jimmidyson:
Thanks @richm!

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -71,20 +71,20 @@ def syms_to_strs(hsh)
...
         annotations = match_annotations(syms_to_strs(metadata['metadata']['annotations'].to_h))
         if @de_dot
           self.de_dot!(labels)
+          self.de_dot!(annotations)
         end
         kubernetes_metadata = {
             'namespace_name' => namespace_name,
             'pod_id'         => metadata['metadata']['uid'],
             'pod_name'       => pod_name,
-            'container_name' => container_name,
```
ewolinetz: even if we aren't using the container name for metadata, it would still be useful to have the container name as a log field

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -244,6 +241,8 @@ def filter_stream_from_files(tag, es)
         end
       end
 
+      metadata['kubernetes']['container_name'] = match_data['container_name'] if match_data
```
jimmidyson: Fixed in #45

richm: Sorry - I thought I was only approving individual commits, not the entire PR.  Not used to the github approval process yet.

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -366,32 +364,22 @@ def start_watch
       watcher.each do |notice|
         case notice.type
           when 'MODIFIED'
-            if notice.object.status.containerStatuses
```
jimmidyson: TBH it probably wasn't filtering much before - containerStatuses are sent with the object anyway. I guess it could filter out pods that are created but have no containers running. OK will add this back in.

jimmidyson: Thinking about it again, I think it's probably reasonable to keep as this does it now, relying on the cache key rather than a field that will almost always be set AFAIK. We can always look at it again if we see problems with it - I haven't, but I'm obviously not running large scale performance tests.

richm: Ok - yeah, if the field is always set it doesn't make much sense to have a condition for it.

PR #51, opened 2016-12-12
Fixed use example on README.md

andor-pierdelacabeza:
The use example on the README file asks to filter kubernetes.var.lib.docker.containers.*.*.log. That filter is wrong, the correct one is kubernetes.var.log.containers.*.log.
That path correlates to a docker install, which by default uses the path /var/lib/docker/containers/[hash]/[hash]-json.log, but in a kubernetes install the correct path is /var/log/containers/*.log


jimmidyson:
Doh! Thanks!

PR #57, opened 2017-02-15
thread safety fix: do not mutate cached object

baconz:
By setting 'container_name' on the metadata object returned by
`@cache.getset`, we create a race condition where containers
in the same pod will end up with an incorrect 'container_name'
assigned.

In the old code `metadata['kubernetes']` was effectively shared
for all records within a pod, so the 'container_name' that ends
up in the final record was determined by whichever record happened
to set 'container_name' last.

Instead, merge the cached object into the metadata that we create
for each record.


portante:
Seems like a reasonable change.  What is holding this up?

jimmidyson:
Thanks @portante for reminding me about this. Sorry this slipped through the net... Stretched a bit thin right now... All help appreciated!

LGTM - just waiting on green then will merge. Thanks @baconz!

PR #59, opened 2017-02-27
Fix config example

### LINKED ISSUE #58: Pods with dot in the pod_name don't get metadata their added
Unless i'm missing something, the fix in b7ca287752585f03fc4188d2f69897e1c95814d8 is incomplete.

The regex is now correct (manually verified) but for some unknown reason logs from pods named with a dot anywhere in the name still don't get collected.

I used [grokdebug](http://grokdebug.herokuapp.com/) to check the regex, and it seems OK for my pod names.
Regex (default): `var\.log\.containers\.(?<pod_name>[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*)_(?<namespace>[^_]+)_(?<container_name>.+)-(?<docker_id>[a-z0-9]{64})\.log$`

Example pod tags:
```
kubernetes.var.log.containers.heapster-v1.2.0-4088228293-11d9f_kube-system_heapster-nanny-7f3832b97b974a41e3ef94c828de5e564c2df0bd4a291156b9f9c05ef004832c.log
kubernetes.var.log.containers.ptlc.debug_default_alpine-e067228f5c2d8f061727bf0d09f3f70da8a5356afd18eb1b6f1b6897c0b0f126.log
kubernetes.var.log.containers.kube-apiserver-ip-10-50-45-119.eu-west-1.compute.internal_kube-system_kube-apiserver-c3e78cbaa087b722ae5d72cbbd9ed25a60e93febf9333b750da0b48e8300b502.log
```

This is impacting deployment schemes that include the hostname in the kubernetes master component pod names, or use version numbering in semver format like the heapster deployment.

How do reproduce:
Deploy a test pod with a dot in the name:
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: fluentd.debug
  labels:
    app: debug
spec:
  containers:
  - command:
    - /bin/sh
    - -c
    - "while true; do date; sleep 5;done"
    image: "alpine:3.4"
    imagePullPolicy: IfNotPresent
    name: alpine
```

Observe missing metadata tags for this pod by querying your logs for entries without the metadata (eg `NOT _exists_: kubernetes` for kibana/elasticsearch)

My hunch is that the metadata is silently dropped here https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/blob/master/lib/fluent/plugin/filter_kubernetes_metadata.rb#L76

pieterlange:
The previous example would skip containers with a dot in the name.

Fixes #58
Followup on https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/issues/35

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ Reading from the JSON formatted log files with `in_tail` and wildcard filenames:
   read_from_head true
 </source>
 
-<filter kubernetes.var.log.containers.*.log>
+<filter kubernetes.var.log.containers.**.log>
   type kubernetes_metadata
 </filter>
 


jimmidyson:
Oh thanks! Didn't know that... Just found http://docs.fluentd.org/v0.12/articles/config-file#match-pattern-how-you-control-the-event-flow-inside-fluentd which explains it too.

Merging...

pieterlange:
Thanks! Hope this saves someone else a couple of hours of nearly going insane :laughing:

jimmidyson:
:100: I'm sure it will!

PR #63, opened 2017-04-06
Assign container_name regardless of @kubernetes_url?

Bekt:
Correct me if I'm wrong, but `container_name` comes from the tag, thus should work even if kubernetes_url is not available.


Bekt:
Apologies, I misread the code. This is already outside of the `@kubernetes_url.present?` block, so everything is already working as expected. 

However, my change looks slightly cleaner IMO. I can change the commit message and PR title and push again if you though.

PR #71, opened 2017-04-27
Let user know more on kubernetes comminication error

shagabutdinov:
I've been installing fluentd to kubernetes and got error 403 without further explanation of what is wrong. Had to build own image in order to debug this behaviour... 

Here is corresponding fluentd issue: https://github.com/fluent/fluentd-kubernetes-daemonset/issues/14


jimmidyson:
Thanks @shagabutdinov!

PR #72, opened 2017-04-30
fix journal CONTAINER_NAME pattern

richm:
The default pattern used to match CONTAINER_NAME was too restrictive,
and did not match the official format defined here:
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/dockertools/docker.go#L317
This patch updates to the correct parse format.


richm:
@jcantrill @ewolinetz PTAL
logging && logging CI is broken without this fix

PR #87, opened 2017-10-20
Add behavioral stats and better logging

portante:

PR #89, opened 2017-10-27
Add stats to cache and watch access

jcantrill:
This PR adds stat info regarding cache hits to superced: #87 

cc @portante

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,6 @@
 source 'https://rubygems.org'
 
-gem 'codeclimate-test-reporter', :group => :test, :require => nil
+gem 'codeclimate-test-reporter', '<1.0.0', :group => :test, :require => nil
 gem 'rubocop', require: false
 
 # Specify your gem's dependencies in fluent-plugin-add.gemspec
--- a/lib/fluent/plugin/filter_kubernetes_metadata.rb
+++ b/lib/fluent/plugin/filter_kubernetes_metadata.rb
@@ -16,6 +16,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+require_relative 'kubernetes_metadata_stats'
 module Fluent
   class KubernetesMetadataFilter < Fluent::Filter
     K8_POD_CA_CERT = 'ca.crt'
@@ -57,6 +58,7 @@ module Fluent
                  :default => '^(?<name_prefix>[^_]+)_(?<container_name>[^\._]+)(\.(?<container_hash>[^_]+))?_(?<pod_name>[^_]+)_(?<namespace>[^_]+)_[^_]+_[^_]+$'
 
     config_param :annotation_match, :array, default: []
+    config_param :stats_interval, :integer, default: 30
 
     def syms_to_strs(hsh)
       newhsh = {}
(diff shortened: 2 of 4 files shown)


Gemfile
```diff
@@ -1,6 +1,6 @@
 source 'https://rubygems.org'
 
-gem 'codeclimate-test-reporter', :group => :test, :require => nil
+gem 'codeclimate-test-reporter', '=0.6.0', :group => :test, :require => nil
```
richm: why not just '< 1.0.0'

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -16,6 +16,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+require File.dirname(__FILE__) + '/kubernetes_metadata_stats'
```
richm: Use http://ruby-doc.org/core-2.0.0/Kernel.html#method-i-require_relative instead?

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -110,9 +112,52 @@ def parse_namespace_metadata(namespace_object)
     def get_pod_metadata(namespace_name, pod_name)
       begin
         metadata = @client.get_pod(pod_name, namespace_name)
-        return if !metadata
-        return parse_pod_metadata(metadata)
-      rescue KubeException
+        if !metadata
```
richm: `unless` instead of `if !`?

richm: or you could combine lines 114 and 115:
```ruby
unless (metadata = @client.get_pod(pod_name, namespace_name))
```

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -110,9 +112,52 @@ def parse_namespace_metadata(namespace_object)
...
+      @cur_time = Time.now
+      return if @curr_time - @prev_time < @stats_interval
+      @prev_time = @curr_time
+      log.info(@stats)
+    end
+
+    def get_namespace_metadata(namespace_name)
+      begin
+        metadata = @client.get_namespace(namespace_name)
+        if !metadata
```
richm:
```ruby
unless (metadata = @client.get_namespace(namespace_name))
```

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -110,15 +112,61 @@ def parse_namespace_metadata(namespace_object)
...
+        log.debug "Exception encountered fetching pod metadata from Kubernetes API #{@apiVersion} endpoint #{@kubernetes_url}: #{e.message}"
+        nil
+      end
+    end
+
+    def dump_stats
+      @curr_time = Time.now
+      return if (@curr_time - @prev_time).to_i < @stats_interval
+      @prev_time = @curr_time
+      log.info(@stats)
```
portante: Can we also dump info about the caches themselves?  How many entries currently maybe?

lib/fluent/plugin/kubernetes_metadata_stats.rb
```diff
@@ -0,0 +1,41 @@
+#
+# Fluentd Kubernetes Metadata Filter Plugin - Enrich Fluentd events with
+# Kubernetes metadata
+#
+# Copyright 2015 Red Hat, Inc.
```
portante: Copyright 2017?

test/plugin/test_cache_stats.rb
```diff
@@ -0,0 +1,39 @@
+#
+# Fluentd Kubernetes Metadata Filter Plugin - Enrich Fluentd events with
+# Kubernetes metadata
+#
+# Copyright 2015 Red Hat, Inc.
```
portante: Copyright 2017?

lib/fluent/plugin/kubernetes_metadata_stats.rb
jcantrill: Most likely can not update this as the other lint checks usually fail when I have tried to update the year

test/plugin/test_cache_stats.rb
jcantrill: Most likely can not update this as the other lint checks usually fail when I have tried to update the year

lib/fluent/plugin/filter_kubernetes_metadata.rb
jcantrill: Not sure what the ask is here.  By dumping the stats you will get a list of all the entries and the associated counts.  What additional value is provided by the raw number you are requesting?

portante: I was thinking we should emit the size of each cache at the least, to know if we are actually evicting on new entries or only on ttl.

PR #90, opened 2017-10-30
WIP - Cache strategies

jcantrill:
@portante this is an initial refactoring to introduce an alternate cache strategy.  I'm thinking we can create an Openshift strategy that relies upon project and namespace UUIDs and also introduces a fallback for the orphans we see

Builds off #89


lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -59,55 +62,11 @@ class KubernetesMetadataFilter < Fluent::Filter
 
     config_param :annotation_match, :array, default: []
     config_param :stats_interval, :integer, default: 10
+    config_param :namespace_watch_strategy, :string, default: 'default'
+    config_param :pod_watch_strategy, :string, default: 'default'
```
richm: there is an `enum` type - see https://github.com/ViaQ/fluent-plugin-viaq_data_model/blob/master/lib/fluent/plugin/filter_viaq_data_model.rb#L113 e.g.
```ruby
   config_param :namespace_watch_strategy, :enum, list: [:default], default: :default
```
the value is a sym, not a string - use it like this
```ruby
  if @namespace_watch_strategy == :default
    ...
```

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -59,55 +62,11 @@ class KubernetesMetadataFilter < Fluent::Filter
 
     config_param :annotation_match, :array, default: []
     config_param :stats_interval, :integer, default: 10
+    config_param :namespace_watch_strategy, :string, default: 'default'
```
portante: Can we call this a "cache" strategy instead of a watch strategy?  Or is this really about watch strategies?  If the are going to call this a catch strategy, then the methods of the cache needs to be hidden so that they can be easily replaced.  The API calls are not covered here, right?

jcantrill:
cc @vvk @r4j4h @boeboe  @overip I would be interested in your input here regarding general use in the community.  This change modifies the caching impl based upon what we have seen because:

* previous cache only used name which does not account for issues where namespaces are deleted and recreated

vvk:
In my experience, I used the plugin to track annotation presence on the namespaces in OpenShift, since namespace = OpenShift project: if the annotation is present on the namespace then applications in the project are logged, if not - aren't logged. The annotation could be added/deleted/changed in runtime at any moment to enable/disable logging, so I need to always track it and see the changes.
I'm not sure whether namespace UUID's ever changed in my scenario, since the namespace wan't ever deleted or recreated, but it's a good point to start tracking it in the plugin.

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -224,40 +232,49 @@ def configure(conf)
...
-              namespace = @client.get_namespace(namespace_name)
-              if namespace
-                parse_namespace_metadata(namespace)
-              end
-            rescue KubeException
-              nil
+    def get_pod_metadata(key, namespace_name, pod_name)
+        ids = @id_cache.getset(key) do 
+          @stats.bump(:pod_cache_miss)
+          pod_metadata = fetch_pod_metadata(namespace_name, pod_name)
```
portante: If `fetch_pod_metadata` returns an empty hash, `{}`, e.g. from lines 83 or 89, will the fetch of `pid_id` at line 257 below return a nil value always?

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -224,40 +232,49 @@ def configure(conf)
...
+          namespace_metadata = fetch_namespace_metadata(namespace_name)
```
portante: If the `pod_metadata` is returned as an empty hash, do we want to always issue the fetch of the namespace name?

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -61,7 +61,7 @@ class KubernetesMetadataFilter < Fluent::Filter
     # parse format is defined here: https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/dockertools/docker.go#L317
     config_param :container_name_to_kubernetes_regexp,
                  :string,
-                 :default => '^(?<name_prefix>[^_]+)_(?<container_name>[^\._]+)(\.(?<container_hash>[^_]+))?_(?<pod_name>[^_]+)_(?<namespace>[^_]+)_(?<pod_uid>[^_])+_[^_]+$'
+                 :default => '^(?<name_prefix>[^_]+)_(?<container_name>[^\._]+)(\.(?<container_hash>[^_]+))?_(?<pod_name>[^_]+)_(?<namespace>[^_]+)_[^_]+_[^_]+$'
```
richm: This field isn't the pod_uid?  What is it?

jcantrill:
closing in favor of #90

PR #93, opened 2017-11-07
refactor cache to use different keys

jcantrill:
The PR refactors the cache impl to:

* Use pod and namespace uid
* handle orphaned records when applicable

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,14 @@
 [![Code Climate](https://codeclimate.com/github/fabric8io/fluent-plugin-kubernetes_metadata_filter/badges/gpa.svg)](https://codeclimate.com/github/fabric8io/fluent-plugin-kubernetes_metadata_filter)
 [![Test Coverage](https://codeclimate.com/github/fabric8io/fluent-plugin-kubernetes_metadata_filter/badges/coverage.svg)](https://codeclimate.com/github/fabric8io/fluent-plugin-kubernetes_metadata_filter)
 
+The Kubernetes metadata plugin filter enriches container log records with pod and namespace metadata.
+
+This plugin derives basic metadata about the container that emitted a given log record using the source of the log record. Records from journald provide metadata about the
+container environment as named fields. Records from JSON files encode metadata about the container in the file name.  The initial metadata derived from the source is used
+to lookup additional metadata about the container's associated pod and namespace (e.g. UUIDs, labels, annotations) when the kubernetes_url is configured.  If the plugin cannot
+authoritatively determine the namespace of the container emitting a log record, it will use an 'orphan' namespace ID in the metadata. This behaviors supports multi-tenant systems
+that rely on the authenticity of the namespace for proper log isolation.
+
 ## Installation
 
     gem install fluent-plugin-kubernetes_metadata_filter
@@ -33,6 +41,10 @@ This must used named capture groups for `container_name`, `pod_name` & `namespac
 * `include_namespace_metadata` - Collect metadata about namespace such as uid, labels and annotations (default: `false`).  
   Note that when this option is enabled, all label from namespace are collected, but annotations are collected according to `annotation_match` value. If `annotation_match` is not presented, no annotations will be collected from the namespace.
 * `annotation_match` - Array of regular expressions matching annotation field names. Matched annotations are added to a log record.
+* `allow_orphans` - Modify the namespace and namespace id to the values of `orphaned_namespace_name` and `orphaned_namespace_id`
+when true (default: `true`)
+* `orphaned_namespace_name` - The namespace to associate with records where the namespace can not be determined (default: `.orphaned`)
+* `orphaned_namespace_id` - The namespace id to associate with records where the namespace can not be determined (default: `orphaned`)
 
 Reading from the JSON formatted log files with `in_tail` and wildcard filenames:
 ```
--- a/lib/fluent/plugin/filter_kubernetes_metadata.rb
+++ b/lib/fluent/plugin/filter_kubernetes_metadata.rb
@@ -2,7 +2,7 @@
 # Fluentd Kubernetes Metadata Filter Plugin - Enrich Fluentd events with
 # Kubernetes metadata
 #
-# Copyright 2015 Red Hat, Inc.
+# Copyright 2017 Red Hat, Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -16,11 +16,19 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+
 require_relative 'kubernetes_metadata_stats'
+require_relative 'kubernetes_metadata_watch_namespaces'
+require_relative 'kubernetes_metadata_watch_pods'
+
 module Fluent
   class KubernetesMetadataFilter < Fluent::Filter
     K8_POD_CA_CERT = 'ca.crt'
     K8_POD_TOKEN = 'token'
+
+    include KubernetesMetadata::Common
+    include KubernetesMetadata::WatchNamespaces
+    include KubernetesMetadata::WatchPods
 
     Fluent::Plugin.register_filter('kubernetes_metadata', self)
 
@@ -39,8 +47,6 @@ module Fluent
     config_param :bearer_token_file, :string, default: nil
     config_param :merge_json_log, :bool, default: true
     config_param :preserve_json_log, :bool, default: true
-    config_param :include_namespace_id, :bool, default: false
-    config_param :include_namespace_metadata, :bool, default: false
     config_param :secret_dir, :string, default: '/var/run/secrets/kubernetes.io/serviceaccount'
     config_param :de_dot, :bool, default: true
     config_param :de_dot_separator, :string, default: '_'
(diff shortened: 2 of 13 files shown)


jcantrill:
cc @portante

README.md
```diff
@@ -3,6 +3,10 @@
 [![Code Climate](https://codeclimate.com/github/fabric8io/fluent-plugin-kubernetes_metadata_filter/badges/gpa.svg)](https://codeclimate.com/github/fabric8io/fluent-plugin-kubernetes_metadata_filter)
 [![Test Coverage](https://codeclimate.com/github/fabric8io/fluent-plugin-kubernetes_metadata_filter/badges/coverage.svg)](https://codeclimate.com/github/fabric8io/fluent-plugin-kubernetes_metadata_filter)
 
+This plugin uses the input source to determine basic metadata to associate with the record.  When `kubernetes_url` is configured, the plugin attempts to fetch additional metadata about the pod
+and its associated namespace.  The plugin will orphan a record if it unable to determine the pod or namespace metadata.  The namespace will be set to `orphaned_namespace_name` and the recored property
```
richm: Either it should not orphan the record if it cannot determine the pod metadata, or that behavior should be configurable.  For our (o-a-l) use case, we only want to orphan records if we cannot determine the namespace metadata - the pod metadata is not so important.

richm: `recorded`

README.md
```diff
@@ -3,6 +3,10 @@
...
+`orphaned_namespace` is set to the eriginal namespace value.
```
richm: `original`

README.md
```diff
@@ -29,8 +33,10 @@ This must used named capture groups for `container_name`, `pod_name` & `namespac
...
 * `container_name_to_kubernetes_regexp` - The regular expression used to extract the k8s metadata encoded in the journal `CONTAINER_NAME` field (default: `'^(?<name_prefix>[^_]+)_(?<container_name>[^\._]+)(\.(?<container_hash>[^_]+))?_(?<pod_name>[^_]+)_(?<namespace>[^_]+)_[^_]+_[^_]+$'`
-  * This corresponds to the definition [in the source](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/dockertools/docker.go#L317)
+* This corresponds to the definition [in the source](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/dockertools/docker.go#L317)
 * `annotation_match` - Array of regular expressions matching annotation field names. Matched annotations are added to a log record.
+* `orphaned_namespace_name` - The namespace to associate with records where the namespace can not be determined (default: `.orphaned`)
```
richm: note that in the previous (current?) implementation, this becomes an index name like `project..orphaned.orphaned...` - what is the intention here?  What is the desired elasticsearch index name?

lib/fluent/plugin/kubernetes_metadata_common.rb
```diff
@@ -0,0 +1,85 @@
+#
+# Fluentd Kubernetes Metadata Filter Plugin - Enrich Fluentd events with
+# Kubernetes metadata
+#
+# Copyright 2015 Red Hat, Inc.
```
richm: This needs to be 2017 - we need to fix the linter or whatever is preventing us from using the correct date

test/plugin/test_filter_kubernetes_metadata.rb
```diff
@@ -165,13 +173,15 @@ def emit_with_tag(tag, msg={}, config='
             'pod_name'       => 'fabric8-console-controller-98rqc',
             'container_name' => 'fabric8-console-container',
             'namespace_name' => 'default',
+            "namespace_id"=>"898268c8-4a36-11e5-9d81-42010af0194c",
```
richm: use single quotes and line up `=>`

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -140,12 +140,15 @@ def initialize
...
       @stats = KubernetesMetadata::Stats.new
       @prev_time = Time.now
-
     end
 
     def configure(conf)
       super
 
+      def log.trace?
+        level == Fluent::Log::LEVEL_TRACE
```
richm: /applause

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -154,7 +157,6 @@ def configure(conf)
         raise Fluent::ConfigError, "Invalid de_dot_separator: cannot be or contain '.'"
       end
 
-      @cache_mutex = Mutex.new
```
richm: We don't need the mutex because it is a thread safe cache?

README.md
jcantrill: This doesn't really speak to what the desired index name is but to how you wish to identify records that are orphaned.  The defaults are the same as they were previously.  It may be we wish the final index to be something like `project._original.orphaned` or different then the default here.  This just adds flexibility

lib/fluent/plugin/kubernetes_metadata_common.rb
jcantrill: We may wish to do this in a subsequent PR since I believe it will snag all files that do not match unless there is a way to ignore year

README.md
richm: I think you mean `record property`?  Maybe use the word `field` (as in a field of the record) instead of `property`?

README.md
```diff
@@ -35,6 +35,8 @@ This must used named capture groups for `container_name`, `pod_name` & `namespac
 * `container_name_to_kubernetes_regexp` - The regular expression used to extract the k8s metadata encoded in the journal `CONTAINER_NAME` field (default: `'^(?<name_prefix>[^_]+)_(?<container_name>[^\._]+)(\.(?<container_hash>[^_]+))?_(?<pod_name>[^_]+)_(?<namespace>[^_]+)_[^_]+_[^_]+$'`
 * This corresponds to the definition [in the source](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/dockertools/docker.go#L317)
 * `annotation_match` - Array of regular expressions matching annotation field names. Matched annotations are added to a log record.
+* `allow_orphans` - Modify the namespace and namespace id to the values of `orphaned_namespace_name` and `orphaned_namespace_id
```
richm: missing closing backtick here?

jcantrill:
@portante added cache changes to address you concerns regarding additional cache misses

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -293,8 +293,14 @@ def get_pod_metadata(key, namespace_name, pod_name, record_create_time)
...
-        metadata =  @cache[ids[:pod_id]] || {}
-        metadata.merge!(@namespace_cache[ids[:namespace_id]] || {}) if ids[:namespace_id]
+        metadata =  @cache.getset(ids[:pod_id]) do
+            m = fetch_pod_metadata(namespace_name, pod_name)
+            (m.nil? || m.empty?) ? nil : m
+        end || {}
+        metadata.merge!(@namespace_cache.getset(ids[:namespace_id]) do
+           m = fetch_namespace_metadata(namespace_name)
+          (m.nil? || m.empty?) ? nil : m
+        end || {})
```
richm: don't need the `if ids[:namespace_id]` any more?

jcantrill: we will find out... rebuilt and trying to test.  i dont know what the cache does when you give it null

jcantrill: simple test of passing key=nil causes the block to be invoked

jcantrill:
@richm When we desire to merge, any thoughts on bumping this to something like 1.0.0 since its just a gross change in caching from the previous mechanism?

richm:
works for me - I don't know what the versioning convention is

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -274,35 +250,70 @@ def configure(conf)
...
+          return {
+            'orphaned_namespace' => namespace_name,
+            'namespace_name' => @orphaned_namespace_name,
+            'namespace_id' => @orphaned_namespace_id
           }
-          metadata.merge!(namespace_metadata) if namespace_metadata
         end
-      end
-      metadata
+        metadata =  @cache.getset(ids[:pod_id]) do
```
portante: If we have `pod_metadata`, why look this up?  We already know we have set this above.  So we only need to look up in the cache if `pod_metadata` is `nil`, right?  Same is true for `namesapce_metadata` below.

portante: Also, we are returning an empty record if we missing looking up the pod ID, but can't we return at least the pod ID as metadata?

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -274,35 +250,70 @@ def configure(conf)
...
+            @stats.bump(:pod_cache_miss)
+            m = fetch_pod_metadata(namespace_name, pod_name)
+            (m.nil? || m.empty?) ? nil : m
+        end || {}
+        metadata.merge!(@namespace_cache.getset(ids[:namespace_id]) do
+           @stats.bump(:namespace_cache_miss)
+           m = fetch_namespace_metadata(namespace_name)
+          (m.nil? || m.empty?) ? nil : m
```
portante: I don't think we want to cache `nil` if we can't fetch the namespace information.

First, at this point, we already have the fixed container ID to pod ID and namespace ID, so we do know this container's namespace ID.  So we can still return a record with the namespace ID in it.   And for that matter, we can return the namespace name we have, just not in a metadata field that says it "is" the namespace name of that ID.

lib/fluent/plugin/filter_kubernetes_metadata.rb
portante: And we can return the pod name we have in hand, just not as the official pod name, just in another field.

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -274,35 +250,70 @@ def configure(conf)
...
+          if ids[:pod_id].nil?
+             @stats.bump(:pod_cache_orphaned_record)
+             ids = :orphaned
+          else
+             @cache[ids[:pod_id]] = pod_metadata
+             @namespace_cache[ids[:namespace_id]] = namespace_metadata
+          end
+          ids
+        end
+        if ids == :orphaned
```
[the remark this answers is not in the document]
jcantrill: It can not because the block is executed on a cache miss where the entry is the set of ids in the other caches.

lib/fluent/plugin/filter_kubernetes_metadata.rb
jcantrill: This allows us to address the case where metadata is rolled off the cache but the IDs still exist in the ID cache.   Updated to return pod id and namespace id if we still get a miss

richm:
@portante we need to merge this soon - do you have any further objections?

portante:
@richm, I spoke with @jcantrill last week and pointed out one missing cache update, and discussed the notion of how we can avoid a second lookup of the same information in certain situations.

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -274,35 +250,70 @@ def configure(conf)
...
-        if @include_namespace_metadata
-          namespace_metadata = @namespace_cache.getset(namespace_name) {
-            @stats.bump(:namespace_cache_miss)
-            get_namespace_metadata(namespace_name)
+    def get_pod_metadata(key, namespace_name, pod_name, record_create_time)
+        ids = @id_cache.getset(key) do
+          @stats.bump(:id_cache_miss)
+          pod_metadata = fetch_pod_metadata(namespace_name, pod_name)
+          namespace_metadata = fetch_namespace_metadata(namespace_name)
+          ids = {:pod_id=> pod_metadata['pod_id'], :namespace_id => namespace_metadata['namespace_id'] }
```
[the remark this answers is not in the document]
portante: @jcantrill and I spoke about missing the `namespace_cache` update that is done on line 276 in the block from logs 279-280.

lib/fluent/plugin/filter_kubernetes_metadata.rb
portante: @jcantrill, I think these comments still hold.  It does not seem to be the case that `getset()` is the right method here.

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -274,35 +252,74 @@ def configure(conf)
...
+
+    def get_pod_metadata(key, namespace_name, pod_name, record_create_time)
+        ids = @id_cache.fetch(key) do
+          #no metadata found
+          @stats.bump(:id_cache_miss)
+          pod_metadata = fetch_pod_metadata(namespace_name, pod_name)
+          namespace_metadata = fetch_namespace_metadata(namespace_name)
+          ids = {:pod_id=> pod_metadata['pod_id'], :namespace_id => namespace_metadata['namespace_id'] }
+          # pod not found or namespace not found
+          if ids[:pod_id].nil? || ids[:namespace_id].nil?
```
portante: We have four cases of behaviors given the possible results of fetching pod and namespace metadata:

 1. we have pod & namespace metadata
 2. we have pod & not namespace metadata
 3. we have namespace & not pod metadata
 4. we have neither

It is possible due to timing and that the metadata fetch APIs are independent, that case #2 can happen, though not often.  It looks like we are ignoring case 2, but can still return an `ids` variable that has a `pod_id` value with a `nil` `namespace_id` value.  Which is probably bad.

This seems to point at a problem with the Kube APIs, we look up by name and pod, but we don't get back the namespace ID with the pod metadata to allow us to validate which namespace it actually is associated with that namespace (as namespace names can come and go).

lib/fluent/plugin/filter_kubernetes_metadata.rb
portante: These comments above seem to still hold: it seems we need to avoid double API calls in the case of the `fetch` block being invoked above.

lib/fluent/plugin/filter_kubernetes_metadata.rb
jcantrill: In this code path we can never have pod metadata without namespace because the client only turns pod meta in context of the namespace.  I would expect a record in this scenario to end up in the orphaned index.

lib/fluent/plugin/filter_kubernetes_metadata.rb
jcantrill: In this case, getset is the correct use since it will: fetch and cache and provide a  return value.  The previous us was incorrect as you noted since we returned meta or :orphaned which was the wrong use.

lib/fluent/plugin/filter_kubernetes_metadata.rb
portante: @jcantrill, not sure about that; if we have the following timeline

 * t0: fetch_pod_metdata(namespace_name, pod_name)
 * t1: namespace is deleted
 * t2: fetch_namespace_metadata(namespace_name)

Then it seems we have this problem.  If the pod metadata returned the namespace ID, then we'd be able to detect this situation.  Right now, at least from what I understand, we can't.

jcantrill:
Merging with approval from @richm and tacit begrudging approval from @portante

portante:
@jcantrill, here is the gist of what I am thinking we need for the cache algorithm: https://gist.github.com/portante/a5a63e776358144c1d7caf161d5cda01

PR #103, opened 2017-12-08
Modify new cache imple with fast path

jcantrill:
Modify caching based on @portante impl

### DIFF
--- a/lib/fluent/plugin/filter_kubernetes_metadata.rb
+++ b/lib/fluent/plugin/filter_kubernetes_metadata.rb
@@ -17,6 +17,8 @@
 # limitations under the License.
 #
 
+require_relative 'kubernetes_metadata_cache_strategy'
+require_relative 'kubernetes_metadata_common'
 require_relative 'kubernetes_metadata_stats'
 require_relative 'kubernetes_metadata_watch_namespaces'
 require_relative 'kubernetes_metadata_watch_pods'
@@ -26,6 +28,7 @@ module Fluent
     K8_POD_CA_CERT = 'ca.crt'
     K8_POD_TOKEN = 'token'
 
+    include KubernetesMetadata::CacheStrategy
     include KubernetesMetadata::Common
     include KubernetesMetadata::WatchNamespaces
     include KubernetesMetadata::WatchPods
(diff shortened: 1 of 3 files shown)


lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -266,51 +266,70 @@ def get_metadata_for_record(match_data, cache_key, create_time)
...
+          # pod found and namespace found
+          metadata = pod_metadata
+          metadata.merge!(namespace_metadata)
+        else
+          if ids[:pod_id].nil? && !ids[:namespace_id].nil?
+            # pod not found, but namespace found
+            @stats.bump(:id_cache_pod_not_found_namespace)
+            ns_time = Time.parse(namespace_metadata['creation_timestamp'])
+            if ns_time <= record_create_time
+              # namesapce is older then record for pod
```
portante: s/namesapce/namespace/, I spelled it wrong.

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -266,51 +266,70 @@ def get_metadata_for_record(match_data, cache_key, create_time)
...
         end)
-        metadata.delete('creation_timestamp') #remove namespace info thats only used for comparison
-        metadata
+      else
+        # SLOW PATH
+        @stats.bump(:id_cache_miss)
+        pod_metadata = fetch_pod_metadata(namespace_name, pod_name)
+        namespace_metadata = fetch_namespace_metadata(namespace_name)
+        ids = { :pod_id=> pod_metadata['pod_id'], :namespace_id => namespace_metadata['namespace_id'] }
+        if !ids[:pod_id].nil? && !ids[:namespace_id].nil?
```
portante: Thanks for fixing this, I updated the original gist as well.

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -266,51 +266,70 @@ def get_metadata_for_record(match_data, cache_key, create_time)
...
+            @stats.bump(:id_cache_pod_not_found_namespace)
+            ns_time = Time.parse(namespace_metadata['creation_timestamp'])
+            if ns_time <= record_create_time
+              # namesapce is older then record for pod
+              ids[:pod_id] = key
+              metadata = @cache.fetch(ids[:pod_id]) do
+                m = { 'pod_id' => ids[:pod_id] }
+              end
+            end
+            metdata.merge!(namespace_metadata)
```
richm: `metadata`?

lib/fluent/plugin/kubernetes_metadata_cache_strategy.rb
```diff
@@ -0,0 +1,91 @@
...
+        ids = { :pod_id=> pod_metadata['pod_id'], :namespace_id => namespace_metadata['namespace_id'] }
+        if !ids[:pod_id].nil? && !ids[:namespace_id].nil?
+          # pod found and namespace found
+          metadata = pod_metadata
+          metadata.merge!(namespace_metadata)
+        else
+          if ids[:pod_id].nil? && !ids[:namespace_id].nil?
+            # pod not found, but namespace found
+            @stats.bump(:id_cache_pod_not_found_namespace)
+            ns_time = Time.parse(namespace_metadata['creation_timestamp'])
```
richm: it would be better to store this pre-parsed in the metadata e.g.
```ruby
namespace_metadata['creation_timestamp_ts'] = Time.parse(namespace_metadata['creation_timestamp'])
```
then you would need to remove both `creation_timestamp_ts` and `creation_timestamp` after merging the metadata into the record.

test/plugin/test_cache_strategy.rb
```diff
@@ -0,0 +1,169 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+require_relative '../helper'
+require 'fluent/plugin/kubernetes_metadata_cache_strategy'
+require 'fluent/plugin/kubernetes_metadata_stats'
```
richm: use `require_relative` for these too, to facilitate testing by doing `ADD *.rb /etc/fluent/plugin/` in the image.

lib/fluent/plugin/kubernetes_metadata_cache_strategy.rb
```diff
@@ -0,0 +1,91 @@
...
+            if ns_time <= record_create_time
```
richm: Is there a test for this case?

jcantrill: This should be the testcase that satisfies this path: https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/pull/103/files#diff-6c11ec708e3e636a7c0eead24e437db5R128

test/plugin/test_cache_strategy.rb
jcantrill: Updated file to use require_relative and also added to the main class file.
I don't think I need to add any docker statements since this is brought in using the gem file for the plugin.  They will be part of the installation.

lib/fluent/plugin/kubernetes_metadata_cache_strategy.rb
jcantrill: I presume this suggestion is for a performance optimization.  This parse statement happens after the fetch of the namespace metadata and will always be string.  What is the advantage of parsing and adding if we never use again?

richm: Yes, for performance.  Yes, if only used once, then not necessary.

PR #106, opened 2018-01-03
preserve nested log field through merge

zeiler:
This fixes an issue when you have a valid log field with the same name within your json log that is being merged. For example if the merge_json_log_key is "log" and you have a json blob as follows: 
```
{"something": "blah", "log": "I want to preserve this"}
```
then that inner "I want to preserve this" message will be dropped when preserve_json_log  is set. This is not good as the purpose of that flag is whether or not to preserve the outer json, not to eliminate the inner field. This PR properly preserves that inner field after the merge.

### DIFF
--- a/lib/fluent/plugin/filter_kubernetes_metadata.rb
+++ b/lib/fluent/plugin/filter_kubernetes_metadata.rb
@@ -395,9 +395,13 @@ module Fluent
         value = record[@merge_json_log_key].strip
         if value[0].eql?('{') && value[-1].eql?('}')
           begin
-            record = JSON.parse(value).merge(record)
+            unpacked = JSON.parse(value)
+            record = unpacked.merge(record)
             unless @preserve_json_log
               record.delete(@merge_json_log_key)
+              if unpacked.has_key?(@merge_json_log_key)
+                record[@merge_json_log_key] = unpacked[@merge_json_log_key]
+              end
             end
           rescue JSON::ParserError=>e
             @stats.bump(:merge_json_parse_errors)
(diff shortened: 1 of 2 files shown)


jcantrill:
@zeiler do you have a test to go along with this change?

zeiler:
Unfortunately I didn't add a test originally because I could not run this repo because of the following error: 
```
An error occurred while installing charlock_holmes (0.7.5), and Bundler cannot continue.
Make sure that `gem install charlock_holmes -v '0.7.5'` succeeds before bundling.
```
Will try relying on the CircleCI tests here.

richm:
there are two problems
- this plugin _never should have had the json parsing logic in it_ :-( it is completely unrelated to the retrieval of k8s metadata - however, we are stuck with it for the time being
- this plugin was never designed to work with a nested `log` field - the problem is that if you want to preserve the original `log` field with the JSON valued string, where do you put it?  I agree that almost all users will want the top level `log` field to contain the value of the nested `log` field, but if you set `preserve_json_log true`, where should the original value go?  We need to have a new top level field e.g. `original_log` to store the old value but we can't be sure if that will conflict with some other field.

richm:
@zeiler can you just disable the `merge_json_log`, and instead use filter_parser?  https://docs.fluentd.org/v0.12/articles/filter_parser

zeiler:
Yeah we actually already got around it by changing our codebase to not emit messages with the special “log” field and then added a catch-all in fluent to rename all other default “log” fields that are not json over to that new field name. We can close this one for now.

PR #65, opened 2017-04-07
Migrate to use v0.14 API

cosmo0920:
If fluentd plugins use and inherits its class v0.14 API, Fluentd permits to send nanosecond precision time and append time zone information.
If this patches are acceptable for this plugin developers, please review carefully and follow this upgrading from v0.12 (old style) guideline: http://docs.fluentd.org/v0.14/articles/plugin-update-from-v12

Kind regards,

PR #111, opened 2018-03-02
FluentD 1.x API and some logging

arcivanov:
Ruby 2.4.3 added to CI
FluentD 1.x API and dependencies
Debug/trace logging for some visibility into initialization

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -7,9 +7,11 @@ dependencies:
     - 'rvm-exec 2.1.9 bundle install'
     - 'rvm-exec 2.2.5 bundle install'
     - 'rvm-exec 2.3.1 bundle install'
+    - 'rvm-exec 2.4.3 bundle install'
 
 test:
   override:
     - 'rvm-exec 2.1.9 bundle exec rake test'
     - 'rvm-exec 2.2.5 bundle exec rake test'
     - 'rvm-exec 2.3.1 bundle exec rake test'
+    - 'rvm-exec 2.4.3 bundle exec rake test'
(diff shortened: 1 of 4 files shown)


richm:
Unfortunately, we still have to support ruby 2.0 for some time due to RHEL7 and CentOS7 dependencies.
I'm not sure how other ruby packages handle support for ruby 2.0 and 2.4 at the same time, which I guess would include support for fluentd 0.12 and 1.0 at the same time.
Do we need a fluentd-1.x branch?  And have separate releases?

arcivanov:
We need fluentd-1.x branch, hence this has been moved to another major version. Usually Fluent plugins have pre-0.14 and post-0.14 versions simultaneously.

https://docs.fluentd.org/v1.0/articles/faq#what-version-of-ruby-does-fluentd-support?
> Fluentd v1.0 works on Ruby 2.1 or later. See v0.12 document for earlier versions.

https://docs.fluentd.org/v1.0/articles/plugin-update-from-v0.12#5.-add-requirements-section
## Requirements

| fluent-plugin-may_awesome | Fluentd     | Ruby   |
|---------------------------|-------------|--------|
| >= 1.0.0                  | >= v0.14.0  | >= 2.1 |
| < 1.0.0                   | >= v0.12.0  | >= 1.9 |

arcivanov:
Alternatively 0.12 version can be frozen for posterity with fixes only and only 1.x+ gets new features. It's really up to the desires of a specific project leadership.

jcantrill:
updated readme in #118

jcantrill:
Does this supercede #65?

Issue #54, opened 2017-01-10
Add docker image to metadata


bcorijn:
Hi,

Would it be possible to add the docker image name:tag to the metadata? It is available from kube pod information and the container_name and container_id are already known from the log. This would allow to trace logs back to specific versions of pods/applications/containers in fast-moving environments, quite a valuable addition in my opinion.

I've looked through the code (I'm not a very experienced ruby developer) and it looks to me like this could be added without too much issues, but I'm not familiar enough to take a stab without getting some feedback where and how this would be best implemented..

mblaschke:
Created a pull request with this feature.
The code is fetching the information from `status.containerStatuses` and uses the container ID for lookup of the information.

Informations published:
- `kubernetes.container_image`
- `kubernetes.container_image_id`

If the information cannot be found (caching issues or what ever happend) the image will be logged as `unknown`.

PR #112, opened 2018-03-10
Implement kubernetes.container_image

mblaschke:
Implementing adding of `kubernetes.container_image` and `kubernetes.container_image_id` from `status.containerStatuses` information as requested in #54

There are no additional api calls and this change is using the existing caching.

First version: get info from pod spec (can cause caching issues)
Second version: get from containerStatuses (prevents possible caching issues, better to display unknown docker image instead of possible wrong docker image)

Available as docker image: `mblaschke/fluentd`

### DIFF
--- a/lib/fluent/plugin/filter_kubernetes_metadata.rb
+++ b/lib/fluent/plugin/filter_kubernetes_metadata.rb
@@ -262,17 +262,25 @@ module Fluent::Plugin
 
     end
 
-    def get_metadata_for_record(match_data, cache_key, create_time, batch_miss_cache)
+    def get_metadata_for_record(match_data, container_id, create_time, batch_miss_cache)
       namespace_name = match_data['namespace']
       pod_name = match_data['pod_name']
+      container_name = match_data['container_name']
       metadata = {
-        'container_name' => match_data['container_name'],
-        'namespace_name' => namespace_name,
-        'pod_name'       => pod_name
+        'container_name'  => container_name,
+        'namespace_name'  => namespace_name,
+        'pod_name'        => pod_name
       }
       if @kubernetes_url.present?
-        pod_metadata = get_pod_metadata(cache_key, namespace_name, pod_name, create_time, batch_miss_cache)
+        pod_metadata = get_pod_metadata(container_id, namespace_name, pod_name, create_time, batch_miss_cache)
+
+        if (pod_metadata.include? 'containers') && (pod_metadata['containers'].include? container_id)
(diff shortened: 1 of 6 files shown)


lib/fluent/plugin/kubernetes_metadata_common.rb
```diff
@@ -54,10 +54,28 @@ def parse_pod_metadata(pod_object)
...
+      begin
+        pod_object['status']['containerStatuses'].each do|container_status|
+          container_meta[container_status['containerID']] = {
+              'name' => container_status['name'],
+              'image' => container_status['image'],
+              'image_id' => container_status['imageID']
+          }
+        end
+      rescue
+        pod_name = pod_object['metadata']['name']
```
jcantrill: I would recommend wrapping this in a `log.debug?` since the only time you would need to retrieve this info is when logging the debug message.

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -255,17 +255,27 @@ def log.trace?
...
       pod_name = match_data['pod_name']
+      container_name = match_data['container_name']
       metadata = {
-        'container_name' => match_data['container_name'],
-        'namespace_name' => namespace_name,
-        'pod_name'       => pod_name
+        'container_name'  => container_name,
+        'namespace_name'  => namespace_name,
+        'pod_name'        => pod_name,
+        'container_image' => 'unknown'
```
jcantrill: I would consider not defaulting this value given the amount of data we transfer for each message is already large.  Couldn't we consider we dont know what it is if its not there.

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -255,17 +255,27 @@ def log.trace?
...
       }
       if @kubernetes_url.present?
-        pod_metadata = get_pod_metadata(cache_key, namespace_name, pod_name, create_time, batch_miss_cache)
+        pod_metadata = get_pod_metadata(container_id, namespace_name, pod_name, create_time, batch_miss_cache)
+
+        docker_container_id = 'docker://' + container_id
```
jcantrill: What if pods are running on CRIO?  I presume the prefix will not be 'docker'

jcantrill:
@mblaschke  Can you squash and rebase on master or v0.12 as appropriate?

mblaschke:
@jcantrill done

PR #115, opened 2018-04-13
Add _jsonparsefailure to tags field if merge_json_log fails

ferbar:
Add "_jsonparsefailure" to tags array if the logmessage starts with "{" and JSON.parse() fails. This is similar to what logstash does.


richm:
I think it would be much better if you did _not_ use this plugin for the json parsing, and instead did the json parsing in a separate pipeline step, using filter_parser or similar.
It is unfortunate that this plugin, the purpose of which is to perform k8s metadata handling, grew an unrelated feature . . .

PR #122, opened 2018-04-20
Turn off JSON payload merges by default

portante:
Merging JSON payloads into the log metadata by default is not a good
idea because of the potential negative impact it can have on the fields
for Elasticsearch.

For example, imagine each log line emitted by a container looks like
the following:
```
{ "abca072a2mwqpc78": 0 }
```
Where the key is "`abca072a2mwqpc78`" and its value is "`0`".

If the container generates a random value for that key, or perhaps a
number in a monotonically increasing sequence, thousands of fields can
be generated for the combined set of JSON documents indexed into
Elasticsearch.  Without an additional filter that could detect such bad
behaviors, and compare against the existing mapping of a target index,
this can cause major problems for Elasticsearch.  Each new field added
to a mapping in Elasticsearch causes a cluster state transition, which
are very very slow (on the order of seconds), and an commensurate
increase in memory usage to track those fields (which can cause
Elasticsearch to exhaust its HEAP, or at least engage in heavy garbage
collection processing).

Use of this option should only be done if the user knows well what they
are doing.

Additionally, with this commit we are deprecated the use of the option,
and will be removing the option and its functionality in a future
release.  Please consider the use of the fluentd `filter_parser` [1]
instead.

[1] https://docs.fluentd.org/v0.12/articles/filter_parser

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,8 +42,10 @@ This must used named capture groups for `container_name`, `pod_name` & `namespac
 * `cache_size` - size of the cache of Kubernetes metadata to reduce requests to the API server (default: `1000`)
 * `cache_ttl` - TTL in seconds of each cached element. Set to negative value to disable TTL eviction (default: `3600` - 1 hour)
 * `watch` - set up a watch on pods on the API server for updates to metadata (default: `true`)
-* `merge_json_log` - merge logs in JSON format as top level keys (default: `true`)
-* `preserve_json_log` - preserve JSON logs in raw form in the `log` key, only used if the previous option is true (default: `true`)
+* `merge_json_log` - merge logs in JSON format as top level keys (default: `true`, DEPRECATED)
+  * WARNING: The configuration option, `merge_json_log` is deprecated and will be removed in a future release; please use fluent's [`filter_parser`](https://docs.fluentd.org/v0.12/articles/filter_parser) instead to parse arbitrary log message payloads as JSON strings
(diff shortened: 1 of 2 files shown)


richm:
Users are encouraged to use https://docs.fluentd.org/v0.12/articles/filter_parser or similar filters if JSON parsing is needed.

jcantrill:
@portante please add a entry in the README to advise this will be deprecated in future version of the plugin.  I think we should encourage consumers to address this kind of functionality outside the use of this plugin.

richm:
Yes - like - `merge_json_log` is deprecated - please use fluent `filter_parser` instead to parse arbitrary fields as JSON strings.

mmsits-kaefferlein:
Hi,
do you have an example how to use it with filter_parser?
And does this work with stackdriver as well? I am currently running into the issue that log is always textpayload - but it should be jsonpayload :(

jcantrill:
@richm We should consider closing this PR in the spirit of least surprise, and opening a new one:

* to remove json_parsing
* bump the release (e.g. 1.1.x, 2.1.x)

jcantrill:
Closing as superceded by #131

portante:
@jcantrill, the progression seems to be to remove the JSON parsing all at once, with no interim release with the warning?  Where did we do that?

jcantrill:
We provided new versions [1][2] that have a different minor release which will allow use of the formal functionality if needed.

[1] https://rubygems.org/gems/fluent-plugin-kubernetes_metadata_filter/versions/2.1.0
[2] https://rubygems.org/gems/fluent-plugin-kubernetes_metadata_filter/versions/1.1.0

Issue #123, opened 2018-05-02
Race condition on short-lived pods


turettn:
I have a k8s cluster that handles a large numbers of Kubernetes Jobs.  Normally, these take a while, but sometimes the job starts and stops almost immediately, at which point it is removed from the cluster.  When this occurs, the logs are still harvested, but when `get_pod_metadata` hits the k8s API, the pod it's looking for no longer exists.  As a result, most of the metadata is missing.

The plugin has already gotten the metadata for the pod as part of the watcher, but discarded it since nothing had asked for it yet - all it did was bump `:pod_cache_watch_misses`.

I have a fix for this, which I'll submit a pull request for shortly.

PR #124, opened 2018-05-02
Uses an optional environment variable to avoid needless cache misses.

turettn:
I have a k8s cluster that handles a large numbers of Kubernetes Jobs. Normally, these take a while, but sometimes the job starts and stops almost immediately, at which point it is removed from the cluster. When this occurs, the logs are still harvested, but when `get_pod_metadata` hits the k8s API, the pod it's looking for no longer exists. As a result, most of the metadata is missing.

The plugin has already gotten the metadata for the pod as part of the watcher, but discarded it since nothing had asked for it yet - all it did was bump `:pod_cache_watch_misses`.

To solve this problem, I added an environment variable called `K8S_NODE_NAME` (easily auto-populated by k8s) that contains the name of the kubernetes node.  Whenever a watch event arrives, before discarding it because the ID is not currently in the cache, it checks to see if the nodeName on the pod matches the node name in the environment variable.  Assuming it matches, it goes ahead and adds the pod metadata to the cache.

This has several a couple advantages:
1. Fixes issue #123, where short-lived pods don't get queried until they no longer exist.
2. It avoids calling the kubernetes pod API to request data that the watcher already had and discarded.

I wasn't super happy with adding environment variables into the mix, but I couldn't come up with a cleaner way.  If anyone has a better idea, I'm happy to discuss it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,22 @@ Reading from the systemd journal (requires the fluentd `fluent-plugin-systemd` a
 </match>
 ```
 
+## Environment variables for Kubernetes
+
+If the name of the Kubernetes node the plugin is running on is set as
+an environment variable with the name `K8S_NODE_NAME`, it will reduce cache
+misses and needless calls to the Kubernetes API.
+
+In the Kubernetes container definition, this is easily accomplished by:
+
+```yaml
+env:
+- name: K8S_NODE_NAME
+  valueFrom:
+    fieldRef:
+      fieldPath: spec.nodeName
+```
+
 ## Example input/output
 
 Kubernetes creates symlinks to Docker log files in `/var/log/containers/*.log`. Docker logs in JSON format.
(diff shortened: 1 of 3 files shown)


jcantrill:
Seems like you still have issue with the metadata getting rolled off the cache that is possibly resolved by increasing the caches size.  This change provides you with the opportunity to further evaluate watch events presumably for pods that are winding down and ultimately will be deleted. 

It LGTM.  Can you please provide a test to support the change.

turettn:
@jcantrill  You’re absolutely right - there is still a chance that the data will be cached, then discarded before logs arrive.  It's definitely less than before, and hopefully acceptably small - k8s should send a MODIFIED when the pod moves to the `Succeeded` or `Failed` state, which would get it into the cache then.  If the default `cache_size` of `1000` pods (on that one host) are cycled through between then and fluent grabbing the logs, the cluster likely has bigger problems than a couple missing labels.

I’ve added test coverage for my changes.  Let me know if there’s anything else I should do.

Thanks!

jcantrill:
@turettn Do we need to back port this to be based on v0.12 ?

PR #127, opened 2018-05-11
bug 1560170. fix error from blocking pipeline if unable to get to api…

jcantrill:
… server

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1560170

by catching all exceptions

### DIFF
--- a/lib/fluent/plugin/filter_kubernetes_metadata.rb
+++ b/lib/fluent/plugin/filter_kubernetes_metadata.rb
@@ -90,12 +90,12 @@ module Fluent
           rescue Exception=>e
             log.debug(e)
             @stats.bump(:pod_cache_api_nil_bad_resp_payload)
-            log.trace("returning empty metadata for #{namespace_name}/#{pod_name} due to error") if log.trace?
+            log.trace("returning empty metadata for #{namespace_name}/#{pod_name} due to error '#{e}'") if log.trace?
           end
         end
-      rescue KubeException=>e
+      rescue Exception=>e
         @stats.bump(:pod_cache_api_nil_error)
-        log.debug "Exception encountered fetching pod metadata from Kubernetes API #{@apiVersion} endpoint #{@kubernetes_url}: #{e.message}"
+        log.debug "Exception '#{e}' encountered fetching pod metadata from Kubernetes API #{@apiVersion} endpoint #{@kubernetes_url}"
       end
       {}
     end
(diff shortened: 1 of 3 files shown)


lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -93,7 +93,7 @@ def fetch_pod_metadata(namespace_name, pod_name)
             log.trace("returning empty metadata for #{namespace_name}/#{pod_name} due to error") if log.trace?
           end
         end
-      rescue KubeException=>e
+      rescue Exception=>e
         @stats.bump(:pod_cache_api_nil_error)
         log.debug "Exception encountered fetching pod metadata from Kubernetes API #{@apiVersion} endpoint #{@kubernetes_url}: #{e.message}"
```
richm: `"Exception #{e} encountered ....` - when debugging, it is good to know exactly what the exception is - same below

richm (approved):
/lgtm

lib/fluent/plugin/filter_kubernetes_metadata.rb
jcantrill: Updated

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -132,12 +132,12 @@ def fetch_namespace_metadata(namespace_name)
...
             @stats.bump(:namespace_cache_api_nil_bad_resp_payload)
-            log.trace("returning empty metadata for #{namespace_name} due to error") if log.trace?
+            log.trace("returning empty metadata for #{namespace_name} due to error '#{e}'") if log.trace?
           end
         end
-      rescue KubeException => kube_error
+      rescue Exception => kube_error
         @stats.bump(:namespace_cache_api_nil_error)
-        log.debug "Exception encountered fetching namespace metadata from Kubernetes API #{@apiVersion} endpoint #{@kubernetes_url}: #{kube_error.message}"
+        log.debug "Exception '#{e}' encountered fetching namespace metadata from Kubernetes API #{@apiVersion} endpoint #{@kubernetes_url}: #{kube_error.message}"
```
richm: `#{kube_error}`

richm: I don't think we need to print both the exception and exception.message

PR #133, opened 2018-05-16
Add the functionality to enrich the record with metadata of a specific pod

yantang-msft:
I'd like to run fluentd as a sidecar instead of daemonset.
In this case, providing the config parameter of the desired pod will be much easier configuration wise, and also reduce the watch events that could incur.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,10 @@ This must used named capture groups for `container_name`, `pod_name` & `namespac
 when true (default: `true`)
 * `orphaned_namespace_name` - The namespace to associate with records where the namespace can not be determined (default: `.orphaned`)
 * `orphaned_namespace_id` - The namespace id to associate with records where the namespace can not be determined (default: `orphaned`)
+* `<metadata_source> section` - The names of the kubernetes resource as the source of metadata. You can only specify one kubernetes resource here. This is useful if the fluentd is running as a sidecar, where you only want to enrich the metadata of the pod that the sidecar is running in. You can easily pass the the namespace_name and the pod_name from the [Downward API](https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/).
+  * `namespace_name` - The name of the namespace. This is required if `<metadata_source>` is specified.
+  * `pod_name` - The name of the pod. This is required if `<metadata_source>` is specified.
+  * `container_name` - The name of the container. Metadata of the container will be added if `container_name` is specified (default: `nil`)
 
 **NOTE:** As of the release 2.1.x of this plugin, it no longer supports parsing the source message into JSON and attaching it to the
 payload.  The following configuration options are removed:
--- a/README.md
+++ b/README.md
@@ -52,7 +52,8 @@ This must used named capture groups for `container_name`, `pod_name` & `namespac
 when true (default: `true`)
 * `orphaned_namespace_name` - The namespace to associate with records where the namespace can not be determined (default: `.orphaned`)
 * `orphaned_namespace_id` - The namespace id to associate with records where the namespace can not be determined (default: `orphaned`)
-* `<metadata_source> section` - The names of the kubernetes resource as the source of metadata. You can only specify one kubernetes resource here. This is useful if the fluentd is running as a sidecar, where you only want to enrich the metadata of the pod that the sidecar is running in. You can easily pass the the namespace_name and the pod_name from the [Downward API](https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/).
+* `<metadata_source> section` - The names of the kubernetes resource as the source of metadata. You can only specify one kubernetes resource here. This is useful if the fluentd is running as a sidecar, where you only want to enrich the metadata of the pod that the sidecar is running in.  
+You can easily set the the namespace_name and the pod_name through the [Downward API](https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/) as environment variables, and use the [environment varialbes](https://docs.fluentd.org/v1.0/articles/faq#how-can-i-use-environment-variables-to-configure-parameters-dynamically?) to configure the plugin dynamically.
   * `namespace_name` - The name of the namespace. This is required if `<metadata_source>` is specified.
   * `pod_name` - The name of the pod. This is required if `<metadata_source>` is specified.
   * `container_name` - The name of the container. Metadata of the container will be added if `container_name` is specified (default: `nil`)
--- a/lib/fluent/plugin/filter_kubernetes_metadata.rb
+++ b/lib/fluent/plugin/filter_kubernetes_metadata.rb
@@ -259,7 +259,7 @@ module Fluent::Plugin
       end
 
       if @metadata_source
-        log.debug "Will stream given metadata source #{metadata_source}"
+        log.on_debug { log.debug "Will stream given metadata source #{metadata_source}" }
         self.class.class_eval { alias_method :filter_stream, :filter_stream_given_metadata_source }
       elsif @use_journal
         log.debug "Will stream from the journal"
@@ -315,12 +315,12 @@ module Fluent::Plugin
       if @kubernetes_url.present?
         if @pod_metadata.nil?
           @pod_metadata = fetch_pod_metadata(namespace_name, pod_name)
-          log.info "Failed to fetch metadata of pod '#{pod_name}' given metadata source: #{metadata_source}" if @pod_metadata.empty?
+          log.on_info { log.info "Failed to fetch metadata of pod '#{pod_name}' given metadata source: #{metadata_source}" } if @pod_metadata.empty?
         end
 
         if @namespace_metadata.nil?
           @namespace_metadata = fetch_namespace_metadata(namespace_name)
-          log.info "Failed to fetch metadata of namespace '#{namespace_name}' given metadata source: #{metadata_source}" if @namespace_metadata.empty?
+          log.on_info { "Failed to fetch metadata of namespace '#{namespace_name}' given metadata source: #{metadata_source}" } if @namespace_metadata.empty?
         end
 
         kubernetes_metadata.merge!(@pod_metadata)
--- a/lib/fluent/plugin/filter_kubernetes_metadata.rb
+++ b/lib/fluent/plugin/filter_kubernetes_metadata.rb
@@ -20,7 +20,9 @@
 require_relative 'kubernetes_metadata_cache_strategy'
 require_relative 'kubernetes_metadata_common'
 require_relative 'kubernetes_metadata_stats'
+require_relative 'kubernetes_metadata_watch_namespace'
 require_relative 'kubernetes_metadata_watch_namespaces'
+require_relative 'kubernetes_metadata_watch_pod'
 require_relative 'kubernetes_metadata_watch_pods'
 
 require 'fluent/plugin/filter'
@@ -32,8 +34,6 @@ module Fluent::Plugin
 
     include KubernetesMetadata::CacheStrategy
     include KubernetesMetadata::Common
-    include KubernetesMetadata::WatchNamespaces
-    include KubernetesMetadata::WatchPods
 
     Fluent::Plugin.register_filter('kubernetes_metadata', self)
 
@@ -71,6 +71,12 @@ module Fluent::Plugin
     config_param :allow_orphans, :bool, default: true
     config_param :orphaned_namespace_name, :string, default: '.orphaned'
     config_param :orphaned_namespace_id, :string, default: 'orphaned'
+
+    config_section :metadata_source, multi: false do
+      config_param :namespace_name, :string
+      config_param :pod_name, :string
+      config_param :container_name, :string, default: nil
+    end
 
     def fetch_pod_metadata(namespace_name, pod_name)
       log.trace("fetching pod metadata: #{namespace_name}/#{pod_name}") if log.trace?
@@ -171,7 +177,7 @@ module Fluent::Plugin
       # Caches pod/namespace UID tuples for a given container UID.
       @id_cache = LruRedux::TTL::ThreadSafeCache.new(@cache_size, @cache_ttl)
 
-      # Use the container UID as the key to fetch a hash containing pod metadata
+      # Use the pod UID as the key to fetch a hash containing pod metadata
       @cache = LruRedux::TTL::ThreadSafeCache.new(@cache_size, @cache_ttl)
 
       # Use the namespace UID as the key to fetch a hash containing namespace metadata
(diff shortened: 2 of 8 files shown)


jcantrill:
In general, I would prefer the 'mode' you are proposing to be better encapsulated with the use of modules targeted to the desired use case.  There is a fair amount of surgery here which interlaces 'if/else' constructs into the existing logic.  Watches can be configured to specific Openshift resources and namespaces. This means you could write a module to watch only the thing you want and mix it in as needed.  Also, I believe fluent plugins supports the notion of config parameters that are blocks which means they are all either on or off together.

yantang-msft:
@jcantrill Thanks for reviewing the code. I just got some time to get back to this. I added new modules for this new "mode", and there is very little "if/else" now.
Again, I'd like to get some more feedbacks before working on the test cases. 😄

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -171,7 +177,7 @@ def log.trace?
       # Caches pod/namespace UID tuples for a given container UID.
       @id_cache = LruRedux::TTL::ThreadSafeCache.new(@cache_size, @cache_ttl)
 
-      # Use the container UID as the key to fetch a hash containing pod metadata
+      # Use the pod UID as the key to fetch a hash containing pod metadata
```
jcantrill: @richm Is this correct?  The info we infer from the file name is not pod UUID as generated by kubernetes but the container UUID as generated by docker is it not?

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -237,13 +243,25 @@ def log.trace?
         end
 
         if @watch
+          if @metadata_source
+            self.class.send(:include, KubernetesMetadata::WatchNamespace)
```
jcantrill: why are we using `self.class.send` instead of just `include KubernetesMetadata::WatchNamespace`

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -237,13 +243,25 @@ def log.trace?
...
           thread = Thread.new(self) { |this| this.start_pod_watch }
           thread.abort_on_exception = true
           namespace_thread = Thread.new(self) { |this| this.start_namespace_watch }
           namespace_thread.abort_on_exception = true
         end
       end
-      if @use_journal
+
+      if @metadata_source
+        log.debug "Will enrich record with metadata of #{@metadata_source.namespace_name}/#{@metadata_source.pod_name}/#{@metadata_source.container_name}"
```
jcantrill: consider wrapping this in `log.on_debug do end` since you are building a string

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -237,13 +243,25 @@ def log.trace?
...
+        self.class.class_eval { alias_method :filter_stream, :filter_stream_given_metadata_source }
```
jcantrill: use `alias_method` directly

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -272,11 +290,11 @@ def get_metadata_for_record(match_data, container_id, create_time, batch_miss_ca
         'pod_name'        => pod_name
       }
       if @kubernetes_url.present?
-        pod_metadata = get_pod_metadata(container_id, namespace_name, pod_name, create_time, batch_miss_cache)
+        pod_metadata = get_pod_metadata(cache_key, namespace_name, pod_name, create_time, batch_miss_cache)
 
-        if (pod_metadata.include? 'containers') && (pod_metadata['containers'].include? container_id)
-          metadata['container_image'] = pod_metadata['containers'][container_id]['image']
-          metadata['container_image_id'] = pod_metadata['containers'][container_id]['image_id']
+        if (pod_metadata.include? 'containers') && (pod_metadata['containers'].include? container_name)
```
jcantrill: I have concerns changing to `container_name` from `container_id` will cause caching issues, as in maybe uniqueness is no longer guaranteed.

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -298,6 +316,33 @@ def filter_stream(tag, es)
...
+        'container_name' => @metadata_source.container_name
+      }
+      cache_key = get_id_cache_key_given_metadata_source
+      batch_miss_cache = {}
+      metadata = {
+        'kubernetes' => get_metadata_for_record(match_data, cache_key, create_time_from_record(es.first[1]), batch_miss_cache)
+      }
+
+      es.each do |time, record|
+        record = record.merge(Marshal.load(Marshal.dump(metadata))) if metadata
```
jcantrill: Why do we need to serialize and then deserialize the hash here?

lib/fluent/plugin/kubernetes_metadata_common.rb
```diff
@@ -61,10 +61,10 @@ def parse_pod_metadata(pod_object)
         pod_object['status']['containerStatuses'].each do|container_status|
           # get plain container id (eg. docker://hash -> hash)
           container_id = container_status['containerID'].sub /^[-_a-zA-Z0-9]+:\/\//, ''
-          container_meta[container_id] = {
-              'name' => container_status['name'],
-              'image' => container_status['image'],
-              'image_id' => container_status['imageID']
+          container_meta[container_status['name']] = {
```
jcantrill: again.  I have reservations about this change

lib/fluent/plugin/filter_kubernetes_metadata.rb
yantang-msft: I tend to not build a string, but since you asked, I'm building one 😄

lib/fluent/plugin/filter_kubernetes_metadata.rb
yantang-msft: It seems I can't directly it directly
> filter_kubernetes_metadata.rb:273:in `configure': private method `alias_method'

yantang-msft:
@jcantrill @richm @portante I tested the functionality manually on my test app, both new and existing code path, and I do have a list of test cases to add. But I foresee that you may have major concerns so I'd like to get some feedbacks first then add the test cases, so I won't waste time on fixing test cases.

As for the new functionality itself, I won't say it's a very narrow use case. It's designed for the sidecar mode, which is a very valid user scenario and a beneficial additive to the existing plugin.

I thought about creating a new plugin, but for the following reasons, contributing to your existing plugin is a much more appropriate approach to me.
1. I would have a lot of duplicated code in the new plugin (authentication, metadata parse, caching strategy...). Comparing to these, the code added is very little.
2. I won't get the upstream updates and fixes
3. It reduces the confusion when people search for k8s metadata plugins
4. After @jcantrill 's suggestion, this change has very minimal interlacement with existing code.

lib/fluent/plugin/filter_kubernetes_metadata.rb
yantang-msft: In this case, there is no pod metadata and thus no container metadata anyways?

yantang-msft:
@jcantrill @richm @portante I have added test cases. Looking forward to your further feedbacks!

portante:
@yantang-msft, have you considered taking the approach of using the [`record_transformer`](https://docs.fluentd.org/v0.12/articles/filter_record_transformer) plugin?

For example:

```
    <filter **>
      @type record_transformer
      enable_ruby
      <record>
        container_name "{ENV['CONTAINER_NAME']}"
        pod_name "{ENV['POD_NAME']}"
        namespace_name "{ENV['NAMESPACE_NAME']}"
      </record>
    </filter>
```

yantang-msft:
@portante I may need more explanation to understand what you suggested.
Are you saying that I can use the `record_transformer` to add the metadata, and thus don't need to use the k8s-metadata-filter at all?

Two advantages of the k8s-metadata-filter are:
1) It adds a lot more metadata like labels, annotations. And since it have all the metadata, it can become even more flexible by leveraging the [record accessor plugin](https://docs.fluentd.org/v1.0/articles/api-plugin-helper-record_accessor).
2) It also has the watch functionality, which simply can't be achieved by the environment variable

portante:
Why not use the record-transformer *in addition to* the k8s-metadata-filter?

yantang-msft:
@portante The problem is after adding those metadata by the record_transformer, the k8s-metadata-filter still doesn't know where does the record come from since it relies on the parsed result of the tag, and thus I won't get extra metadata like labels, annotations...

Sure I can use the record_transformer to change the tag, but the weaknesses are
1) The current filtering logic requires the container_id to be there, which I can't get from the environment variable. Without the container_id, the [fast path](https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/blob/master/lib/fluent/plugin/kubernetes_metadata_cache_strategy.rb#L25) will never happen, and the plugin will call the API server all the time.
2) It's a bit more performant with this change. i.e, no regex matching, no need to watch the change of all namespaces and pods, but only the one that I care about.
3) Running the fluentd agent as a sidecar is a very valid scenario. And it's just more straightforward to have the plugin support this scenario natively.

What do you think?

portante:
@yantang-msft, can you share the fluent.conf file you are using in your side-car container?

yantang-msft:
Here it is https://github.com/yantang-msft/kubernetes-sidecar-diagnostics/blob/scratch/FluentdAgent/conf/fluent.conf
As you can see I have two inputs, the tag from the http input can be any. Having a configuration like this is cleaner than using a record_transformer plus a custom tag regex exp.

richm:
> @yantang-msft, just curious, but can a side car container read from /var/log/containers?

yes, if the pod is running as `privileged`, or has some other `hostmount` permission.  The fluentd pod almost always runs this way.

> It seems that the real problem is this plugin knows too much: it knows about journald data, it knows about docker json-file name and data.

Agreed.  It would be better if the plugin had no knowledge of filename formats or journald CONTAINER_NAME fields.  Ideally, the user would pass in a `kubernetes` container with one or more fields specified, and tell the plugin to fill in the rest e.g.
```
record = {"kubernetes":
  "pod_name": "mypodname",
  "namespace_name": "mynamespace"
}
```
The plugin would look up the missing metadata and fill in the kubernetes container in the record.

yantang-msft:
@portante @richm  @jcantrill I agree the plugin knows too much, which is a good default config experience though. And I think we are kind of on the same page that this plugin just need to know some key names (namespace, pod, container) and fills in the rest of the metadata.

The only difference between this PR and what @portante suggested is this PR aims at running the plugin as a sidecar, and user can only specify one pod for the plugin. While specifying the key names are more general, the plugin can be run as a daemonset and serve any number of pods.

However, I see a hard problem to support running the plugin as a daemonset, which is described in details below. And I want to know what do you guys think.
Would you consider taking this PR which addresses the need to run the plugin as a sidecar? Or do you want to make it general still? I can't think of a elegant reliable solution, but I'm willing to discuss how good it can be made.

### The Problem
A key problem of running the plugin as a daemonset is you almost always want the `container_id` to be there. Otherwise if we only have namespace_name and pod_name, we would have to use that as the cache key. But namespace and pod name are not as unique as `container_id`, and can be recreated with the same name.

You can turn on watch as a rescue, but there will still be timing issues. For example, the plugin receives record from a new pod (with same name), but the watch is not triggered yet, thus metadata of old pod is attached to record from new pod. Or there is slow processing of record from old pod, but the cache has been updated with new pod's metadata, thus metadata of new pod is attached to record from old pod.

This is not a problem for a sidecar. When the sidecar is running and processing the metadata, it can always get the correct metadata just from the namespace name and pod name.

yantang-msft:
Actually, let me find some time tomorrow and draft the change of supporting the key names, so we can have something concrete for discussion.

yantang-msft:
@portante Exactly! And I think you're right we can totally get rid of the caching logic in this case. I will modify the code so it doesn't call into the metadata_cache_strategy at all.

Thus the sidecar case will become a separate simple code path. Supporting multiple pods will be a another code path. They are both valid scenarios. Having two code paths makes the logic simple and easy to maintain. How about that? It sounds good to me.

yantang-msft:
@portante @richm @jcantrill I implemented the logic in a new way.
Now the sidecar case is pretty straightforward. It only fetches the metadata for once, and the complicated cache_strategy is avoided.

yantang-msft:
@portante @richm @jcantrill A kind reminder of this PR, if you have time.

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -273,18 +291,64 @@ def get_metadata_for_record(match_data, container_id, create_time, batch_miss_ca
...
+          'pod_name' => pod_name
+        }
+      }
+      kubernetes_metadata = metadata['kubernetes']
+      kubernetes_metadata['container_name'] = container_name if container_name
+
+      if @kubernetes_url.present?
+        if @pod_metadata.nil?
+          @pod_metadata = fetch_pod_metadata(namespace_name, pod_name)
+          log.info "Failed to fetch metadata of pod '#{pod_name}' given metadata source: #{metadata_source}" if @pod_metadata.empty?
```
jcantrill: Subtle perf improvement:
```
log.on_info do 
  log.info("the message")
end
... since you are constructing a string in the message

File: README.md [hunk 9]
```diff
@@ -52,6 +52,10 @@ This must used named capture groups for `container_name`, `pod_name` & `namespac
 when true (default: `true`)
 * `orphaned_namespace_name` - The namespace to associate with records where the namespace can not be determined (default: `.orphaned`)
 * `orphaned_namespace_id` - The namespace id to associate with records where the namespace can not be determined (default: `orphaned`)
+* `<metadata_source> section` - The names of the kubernetes resource as the source of metadata. You can only specify one kubernetes resource here. This is useful if the fluentd is running as a sidecar, where you only want to enrich the metadata of the pod that the sidecar is running in. You can easily pass the the namespace_name and the pod_name from the [Downward API](https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/).
+  * `namespace_name` - The name of the namespace. This is required if `<metadata_source>` is specified.
+  * `pod_name` - The name of the pod. This is required if `<metadata_source>` is specified.
```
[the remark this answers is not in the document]
yantang-msft: Yes, user will have to use environment variable and get the pod name through downward APIs to configure this plugin dynamically. This is how I configured it: [fluent config](https://github.com/yantang-msft/kubernetes-sidecar-diagnostics/blob/scratch/FluentdAgent/conf/fluent.conf#L37), [yaml](https://github.com/yantang-msft/kubernetes-sidecar-diagnostics/blob/scratch/TestHttpLogSidecar.yaml)
I updated the document so it should be clear enough.

lib/fluent/plugin/filter_kubernetes_metadata.rb
yantang-msft: Updated.

richm:
Does https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/pull/155 solve this problem?

Issue #146, opened 2018-09-05
OneEventStream throws undefined error method 'empty?'


jcantrill:
```
2018-09-05 19:42:08 +0000 [warn]: emit transaction failed: error_class=NoMethodError error="undefined method `empty?' for #<Fluent::OneEventStream:0x007fcc31e86fc0>" location="/usr/share/gems/gems/fluent-plugin-kubernetes_metadata_filter-1.0.3/lib/fluent/plugin/filter_kubernetes_metadata.rb:287:in `filter_stream_from_files'" tag="journal"
```

jcantrill:
@richm Released in:
* v1.2.0
* v2.1.3

PR #147, opened 2018-09-07
[master] test for eventstream has empty method and if class is EventStream

richm:
Also add an explicit test to see if OneEventStream works


richm:
fixes https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/issues/146 in master

PR #150, opened 2018-09-24
Implement docker_image meta information

josefkarasek:
(cherry picked from commit 456315f5c21ff3db7e8822bc4741f7fe8b9576e1)

PR #151, opened 2018-09-24
[v0.12] detect metadata source

josefkarasek:


josefkarasek:
depends on #150

README.md
```diff
@@ -44,21 +44,33 @@ This must used named capture groups for `container_name`, `pod_name` & `namespac
...
 * `de_dot_separator` - separator to use if `de_dot` is enabled (default: `_`)
-* `use_journal` - If false (default), messages are expected to be formatted and tagged as if read by the fluentd in\_tail plugin with wildcard filename.  If true, messages are expected to be formatted as if read from the systemd journal.  The `MESSAGE` field has the full message.  The `CONTAINER_NAME` field has the encoded k8s metadata (see below).  The `CONTAINER_ID_FULL` field has the full container uuid.  This requires docker to use the `--log-driver=journald` log driver.
+* *DEPRECATED* `use_journal` - If false, messages are expected to be formatted and tagged as if read by the fluentd in\_tail plugin with wildcard filename.  If true, messages are expected to be formatted as if read from the systemd journal.  The `MESSAGE` field has the full message.  The `CONTAINER_NAME` field has the encoded k8s metadata (see below).  The `CONTAINER_ID_FULL` field has the full container uuid.  This requires docker to use the `--log-driver=journald` log driver.  If unset (the default), the plugin will use the `CONTAINER_NAME` and `CONTAINER_ID_FULL` fields
```
jcantrill: @richm is this even `DEPRECATED` if we are 'agnostic' of the source?  Is it needed?

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -253,94 +251,95 @@ def log.trace?
...
+      time_key = @time_fields.detect{ |ii| record.has_key?(ii) }
+      time = record[time_key]
+      if time.nil? || time.chop.empty?
+        return internal_time
       end
-
-      es.each do |time, record|
-        record = record.merge(Marshal.load(Marshal.dump(metadata))) if metadata
-        new_es.add(time, record)
+      if time_key == '_SOURCE_REALTIME_TIMESTAMP' || time_key == '__REALTIME_TIMESTAMP'
```
jcantrill: `if ['_SOURCE_REALTIME_TIMESTAMP','__REALTIME_TIMESTAMP']`.include?(time_key)`

and that array could be reused in the other code bit that pushes them onto an array:  `source.concat(other)`

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -253,94 +251,95 @@ def log.trace?
...
+        if tag_match_data && tag_metadata.nil?
+          tag_metadata = get_metadata_for_record(tag_match_data['namespace'], tag_match_data['pod_name'], tag_match_data['container_name'],
+            tag_match_data['docker_id'], create_time_from_record(record, time), batch_miss_cache)
+        end
+        metadata = Marshal.load(Marshal.dump(tag_metadata)) if tag_metadata
+        if (@use_journal || @use_journal.nil?) &&
+          (j_metadata = get_metadata_for_journal_record(record, time, batch_miss_cache))
+          metadata = j_metadata
+        end
+        if @lookup_from_k8s_field && record.has_key?('kubernetes') && record.has_key?('docker') &&
```
jcantrill: I don't understand why the record would ever have any of the information populated.   We use the various regex to pull the basic info from either the file path or journal and add them to  the record.  I would expect we still need to do that here.  Did we change something such that records now already include namespace, pod name, container id, and container name?  Was there a change to modify a parser?

README.md
```diff
@@ -44,21 +44,33 @@ This must used named capture groups for `container_name`, `pod_name` & `namespac
...
 * `annotation_match` - Array of regular expressions matching annotation field names. Matched annotations are added to a log record.
 * `allow_orphans` - Modify the namespace and namespace id to the values of `orphaned_namespace_name` and `orphaned_namespace_id`
 when true (default: `true`)
 * `orphaned_namespace_name` - The namespace to associate with records where the namespace can not be determined (default: `.orphaned`)
 * `orphaned_namespace_id` - The namespace id to associate with records where the namespace can not be determined (default: `orphaned`)
+* `lookup_from_k8s_field` - If the field `kubernetes` is present, lookup the metadata from the given subfields such as `kubernetes.namespace_name`, `kubernetes.pod_name`, etc.  This allows you to avoid having to pass in metadata to lookup in an explicitly formatted tag name or in an explicitly formatted `CONTAINER_NAME` value.  For example, set `kubernetes.namespace_name`, `kubernetes.pod_name`, `kubernetes.container_name`, and `docker.id` in the record, and the filter will fill in the rest. (default: `true`)
```
jcantrill: Along with my other comment, I don't know under what conditions these fields would ever be populated from an incoming record since this plugin adds them based on the file path or journal info.

README.md
richm: There may be a use case for the user to force `use_journal true` if the user 1) knows for a fact that journal will be used and 2) needs every jif of extra processing

lib/fluent/plugin/filter_kubernetes_metadata.rb
richm: The point is that this plugin should only know how to look up pod and namespace metadata given a pod name, namespace name, container name.  Why should this plugin have deep knowledge of the container log file format or the CONTAINER_NAME format?  That information could be extracted elsewhere, and passed into this filter plugin via `kubernetes.namespace_name`, etc.

README.md
richm: see above

lib/fluent/plugin/filter_kubernetes_metadata.rb
jcantrill: Agree. but thats now how it works.  Does it come in via a parser? ... and its metadata... the most basic metadata available which is the point of the plugin.  is it unreasonable to suggest it figures it out.  or...  again... is that the 'parser' that does that?

richm: The data could come in via a parser - you could use a record_transformer to parse out the fields from the tag or CONTAINER_NAME - in fact we already do something like this in the mux case, where we parse the incoming tag, then convert it to CONTAINER_NAME for the sole purpose of having the metadata in the right format for the k8s filter plugin - https://github.com/openshift/origin-aggregated-logging/blob/master/fluentd/configs.d/input-post-forward-mux.conf#L30

jcantrill: So we are writing this code in anticipation of some 'upstream' bit in the pipeline being able to provide that information?

richm: yes - I would be able to take advantage of it immediately in mux

jcantrill:
published in https://rubygems.org/gems/fluent-plugin-kubernetes_metadata_filter/versions/1.2.1

PR #155, opened 2018-10-02
deprecate use_journal - add lookup_from_k8s_field - detect input

josefkarasek:
Make the plugin able to detect the source of the metadata,
either from the journal with `CONTAINER_NAME` or from in_tail
with the tag.  Add a new source of metadata - if
`lookup_from_k8s_field true`, then the user may pass in
metadata in the fields
`docker.container_id`, `kubernetes.namespace_name`,
`kubernetes.pod_name`, `kubernetes.container_name`

(cherry picked from commit 83436cba93257f34796033594c6d0af5997900c2)


josefkarasek:
@richm @jcantrill It's a cherrypick, but still. please have a look at it. Thank you.

Issue #149, opened 2018-09-17
Persistent cache with key timeout for the k8s metadata cache to deal with restarts and pod deletion.


HackToHell:
Hello,
In the case where there's a very high rate of logs, the push to ES can slow down considerably. In the mean time, if there's a pod created in namespace, then the pod is deleted, the logs for the particular pod is lost because the API server does not have the pod object. 
The caching logic can be modified so that it removed a pod after X amount of time, this also needs to be persistent to the disk because if the pod restarts, we will loose the in memory cache.

I can send a PR for this change, do you have any suggestions/other solutions?

jcantrill:
PR's are always welcome.  There is a balancing act to be considered here with regards to:
* memory usage
* on disk buffer availability
* metadata retention needs.

I would argue that maybe you need to consider expanding the size of your ES cluster if you exceed the limits beyond what is capable with increasing the buffer size [1].  Also, it would be useful if your cache impl could be a drop in replacement for the current LRU cache [2].

[1] https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/blob/master/lib/fluent/plugin/filter_kubernetes_metadata.rb#L41
[2] https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/blob/master/lib/fluent/plugin/kubernetes_metadata_cache_strategy.rb

josephmcasey:
I recently stumbled upon this problem, and it seems like a functionality gap worth filling. Given the amount of time that has passed from the WIP PR that was never finished. Has a new design come about for addressing this issue that would differs from the WIP PR #157 ?

jcantrill:
Nothing has changed with regards to this issue. I'm not certain when this was created in relation to the cache design change, but the current implementation makes an effort to associate pod meta using time stamps if needed. The reality is that if the info was never fetched then really there is nothing more to associate; meta is only available using pod name and namespace. You can always turn off watch, modify the eviction time and raise the cache count for longer cache availability.  There are currently no plans to add another cache impl. PRs are welcome if you believe there is something that fits your usecase. There is an open PR that provides an alternative cache option that could be used as a starting point.

josephmcasey:
> The reality is that if the info was never fetched then really there is nothing more to associate; meta is only available using pod name and namespace.

[...]

PR #157, opened 2018-10-04
[WIP] Feature add checkpoint cache

HackToHell:
Attempts to fix the problem detailed in #149,I have added in a sqlite db to persist the entries, with a checkpoint thread. 
Unit tests have to be added.

### DIFF
--- a/lib/fluent/plugin/checkpoint_metadata_cache.rb
+++ b/lib/fluent/plugin/checkpoint_metadata_cache.rb
@@ -23,8 +23,8 @@ module KubernetesMetadata
     def start_checkpoint
       while @checkpoint_thread_running
         log.trace "check pointing cache to disk" if log.trace?
-        write_cache_to_file
         prune_old_entries
+        write_cache_to_file
         sleep @checkpoint_interval
       end
     end
@@ -55,7 +55,7 @@ module KubernetesMetadata
     end
 
     def prune_old_entries
-      timestamp = Time.now - @cache_ttl
+      timestamp = Time.now - @checkpoint_ttl
       @db.execute "delete from cache where created_at <= ?", timestamp.to_i
       @db.execute "delete from id_cache where created_at <= ?", timestamp.to_i
       @db.execute "delete from namespace_cache where created_at <= ?", timestamp.to_i
--- a/lib/fluent/plugin/checkpoint_metadata_cache.rb
+++ b/lib/fluent/plugin/checkpoint_metadata_cache.rb
@@ -1,0 +1,89 @@
+#
+# Fluentd Kubernetes Metadata Filter Plugin - Enrich Fluentd events with
+# Kubernetes metadata
+#
+# Copyright 2017 Red Hat, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+require "sqlite3"
+
+module KubernetesMetadata
+  module CheckPointCache
+    def start_checkpoint
+      while true
+        log.trace "check pointing cache to disk" if log.trace?
+        write_cache_to_file
+        prune_old_entries
+        sleep @checkpoint_interval
+      end
+    end
+
+    def initialize_db
+      @db = SQLite3::Database.new @checkpoint_db_path
+      @db.execute <<-SQL
+        create table IF NOT EXISTS cache (
+          id TEXT PRIMARY KEY,
+          val TEXT,
+          created_at INTEGER
+        );
+      SQL
+      @db.execute <<-SQL
+        create table IF NOT EXISTS id_cache (
+          id TEXT PRIMARY KEY,
+          val TEXT,
+          created_at INTEGER
+        );
+      SQL
+      @db.execute <<-SQL
+        create table IF NOT EXISTS namespace_cache (
+          id TEXT PRIMARY KEY,
+          val TEXT,
+          created_at INTEGER
+        );
+      SQL
+    end
+
+    def prune_old_entries
+      timestamp = Time.now - @cache_ttl
+      @db.execute "delete from cache where created_at <= ?", timestamp.to_i
+      @db.execute "delete from id_cache where created_at <= ?", timestamp.to_i
+      @db.execute "delete from namespace_cache where created_at <= ?", timestamp.to_i
+    end
+
+    def write_cache_to_file
+      timestamp = Time.now.to_i
+      @cache.each do |k, v|
+        @db.execute "insert or ignore into cache (id, val, created_at) values ( ?, ?, ?)", [k, JSON.generate(v), timestamp]
+      end
+      @id_cache.each do |k, v|
+        @db.execute "insert or ignore into id_cache (id, val, created_at) values ( ?, ?, ?)", [k, JSON.generate(v), timestamp]
+      end
+      @namespace_cache.each do |k, v|
+        @db.execute "insert or ignore into namespace_cache (id, val, created_at) values ( ?, ?, ?)", [k, JSON.generate(v), timestamp]
+      end
+    end
+
+    def read_cache_from_file
+      @db.execute("select * from cache") do |row|
+        @cache[row[0]] = JSON.parse(row[1])
+      end
+      @db.execute("select * from id_cache") do |row|
+        @id_cache[row[0]] = JSON.parse(row[1])
+      end
+      @db.execute("select * from namespace_cache") do |row|
+        @namespace_cache[row[0]] = JSON.parse(row[1])
+      end
+    end
+  end
+end
--- a/lib/fluent/plugin/checkpoint_metadata_cache.rb
+++ b/lib/fluent/plugin/checkpoint_metadata_cache.rb
@@ -63,16 +63,34 @@ module KubernetesMetadata
 
     def write_cache_to_file
       current_timestamp = Time.now
-      @cache.each do |k, v|
-        @db.execute "insert or ignore into cache (id, val, created_at) values ( ?, ?, ?)", [k, JSON.generate(v), current_timestamp.to_i]
+      begin
+        @db.transaction
+        @cache.each do |k, v|
+          @db.execute "insert or ignore into cache (id, val, created_at) values ( ?, ?, ?)", [k, JSON.generate(v), current_timestamp.to_i]
+        end
+        @db.commit
+      rescue SQLite3::Exception => e
+        log.trace "unable to write into cache due to to error '#{e}'"
       end
-      @id_cache.each do |k, v|
-        @db.execute "insert or ignore into id_cache (id, val, created_at) values ( ?, ?, ?)", [k, JSON.generate(v), current_timestamp.to_i]
+      begin
+        @db.transaction
+        @id_cache.each do |k, v|
+          @db.execute "insert or ignore into id_cache (id, val, created_at) values ( ?, ?, ?)", [k, JSON.generate(v), current_timestamp.to_i]
+        end
+        @db.commit
+      rescue SQLite3::Exception => e
+        log.trace "unable to write into cache due to to error '#{e}'"
       end
-      @namespace_cache.each do |k, v|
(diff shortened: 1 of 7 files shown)


lib/fluent/plugin/checkpoint_metadata_cache.rb
```diff
@@ -0,0 +1,92 @@
...
+      timestamp = Time.now - @checkpoint_ttl
+      @db.execute "delete from cache where created_at <= ?", timestamp.to_i
+      @db.execute "delete from id_cache where created_at <= ?", timestamp.to_i
+      @db.execute "delete from namespace_cache where created_at <= ?", timestamp.to_i
+    end
+
+    def write_cache_to_file
+      current_timestamp = Time.now
+      @cache.each do |k, v|
+        @db.execute "insert or ignore into cache (id, val, created_at) values ( ?, ?, ?)", [k, JSON.generate(v), current_timestamp.to_i]
```
jcantrill: I'm unfamiliar with the sqlite syntax and its transactional nature, but might it be more performant to perform a bulk insert?

lib/fluent/plugin/checkpoint_metadata_cache.rb
```diff
@@ -0,0 +1,92 @@
...
+        sleep @checkpoint_interval
+      end
+    end
+
+    def initialize_db
+      @db = SQLite3::Database.new @checkpoint_db_path
+      @db.execute <<-SQL
+        create table IF NOT EXISTS cache (
+          id TEXT PRIMARY KEY,
+          val TEXT,
```
jcantrill: Is text sufficiently large enough to hold a well formed json blob?

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -69,6 +73,11 @@ class KubernetesMetadataFilter < Fluent::Filter
     config_param :allow_orphans, :bool, default: true
     config_param :orphaned_namespace_name, :string, default: '.orphaned'
     config_param :orphaned_namespace_id, :string, default: 'orphaned'
+    config_param :checkpoint_enabled, :bool, default: true
+    config_param :checkpoint_interval, :integer, default: 30
+    # This is used to prune older entries that will be written to the sqlite db, the most recent entries(aka the hot LRU cache) will always be in the db.
+    config_param :checkpoint_ttl, :integer, default: 3000
+    config_param :checkpoint_db_path, :string, default: '/var/log/k8s-metadata-checkpoint.db'
```
jcantrill: wouldnt a more appropriate path that meets convention be something like `/var/lib/fluent-plugin-kubernetes_metadata_filter/checkpoint.db` or `/var/lib/fluent/k8s-metadata-checkpoint.db`

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -226,20 +236,25 @@ def log.trace?
         rescue KubeException => kube_error
           raise Fluent::ConfigError, "Invalid Kubernetes API #{@apiVersion} endpoint #{@kubernetes_url}: #{kube_error.message}"
         end
+        if @checkpoint_enabled
+          @checkpoint_thread_running = true
```
jcantrill: Shouldnt this come after the the thread initialization in the following lines? maybe need some sort of mutex to ensure it happens only once?

lib/fluent/plugin/kubernetes_metadata_cache_strategy.rb
```diff
@@ -24,12 +24,12 @@ def get_pod_metadata(key, namespace_name, pod_name, record_create_time, batch_mi
       if !ids.nil?
         # FAST PATH
         # Cache hit, fetch metadata from the cache
-        metadata = @cache.fetch(ids[:pod_id]) do
+        metadata = @cache.fetch(ids['pod_id']) do
```
jcantrill: why are we using a string and not the symbol?

lib/fluent/plugin/kubernetes_metadata_cache_strategy.rb
```diff
@@ -24,12 +24,12 @@ def get_pod_metadata(key, namespace_name, pod_name, record_create_time, batch_mi
...
           @stats.bump(:pod_cache_miss)
           m = fetch_pod_metadata(namespace_name, pod_name)
           (m.nil? || m.empty?) ? {'pod_id'=>ids[:pod_id]} : m
         end
-        metadata.merge!(@namespace_cache.fetch(ids[:namespace_id]) do
+        metadata.merge!(@namespace_cache.fetch(ids['namespace_id']) do
```
jcantrill: why are we using a string and not the symbol?

lib/fluent/plugin/checkpoint_metadata_cache.rb
HackToHell: Yup, will do so.

lib/fluent/plugin/checkpoint_metadata_cache.rb
HackToHell: The max value is 2^32 − 1, this _can_ fail in cases with pods with large labels. I will add in a skip for overly large values.

lib/fluent/plugin/filter_kubernetes_metadata.rb
HackToHell: Sure, let me change that to `/var/lib/fluent-plugin-kubernetes_metadata_filter/checkpoint.db`

lib/fluent/plugin/checkpoint_metadata_cache.rb
HackToHell: Switched to transactions, the build insert format had a lot of edge cases and performance wins are minimal.

lib/fluent/plugin/kubernetes_metadata_cache_strategy.rb
HackToHell: Rolled back the change.

HackToHell:
@jcantrill Updated with review comments! I also had to refractor the caching pathway because sometimes it would enter into a loop where the ids would be present in the id cache but not the other two caches, I could not figure out which case caused it to enter into that state, but with the persistent cache, the fluentd plugin would slow down the entire pipeline until the cache was flushed.

lib/fluent/plugin/checkpoint_metadata_cache.rb
```diff
@@ -0,0 +1,111 @@
...
+      @db.execute "delete from id_cache where created_at <= ?", timestamp.to_i
+      @db.execute "delete from namespace_cache where created_at <= ?", timestamp.to_i
+    end
+
+    def write_cache_to_file
+      current_timestamp = Time.now
+      begin
+        @db.transaction
+        @cache.each do |k, v|
+          @db.execute "insert or ignore into cache (id, val, created_at) values ( ?, ?, ?)", [k, JSON.generate(v), current_timestamp.to_i]
```
dhananjaysathe: add a continue to handle the rare cases where len > 2^32 chars case as pointed out above; its okay to skip those imho

lib/fluent/plugin/checkpoint_metadata_cache.rb
```diff
@@ -0,0 +1,111 @@
...
+          @db.execute "insert or ignore into id_cache (id, val, created_at) values ( ?, ?, ?)", [k, JSON.generate(v), current_timestamp.to_i]
+        end
+        @db.commit
+      rescue SQLite3::Exception => e
+        log.trace "unable to write into cache due to to error '#{e}'"
+      end
+      begin
+        @db.transaction
+        @namespace_cache.each do |k, v|
+          @db.execute "insert or ignore into namespace_cache (id, val, created_at) values ( ?, ?, ?)", [k, JSON.generate(v), current_timestamp.to_i]
```
dhananjaysathe: add a continue to handle the rare cases where len > 2^32 chars case as pointed out above; its okay to skip those imho

lib/fluent/plugin/checkpoint_metadata_cache.rb
```diff
@@ -0,0 +1,111 @@
...
+      log.trace "db write took " + (Time.now - current_timestamp).to_s if log.trace?
+    end
+
+    def read_cache_from_file
+      current_timestamp = Time.now
+      @db.execute("select * from cache") do |row|
+        @cache[row[0]] = JSON.parse(row[1])
+      end
+      @db.execute("select * from id_cache") do |row|
+        rv = JSON.parse(row[1])
```
dhananjaysathe: consider using symbolized_names : true while parsing to avoid the manual conversion

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -69,6 +73,11 @@ class KubernetesMetadataFilter < Fluent::Filter
     config_param :allow_orphans, :bool, default: true
     config_param :orphaned_namespace_name, :string, default: '.orphaned'
     config_param :orphaned_namespace_id, :string, default: 'orphaned'
+    config_param :checkpoint_enabled, :bool, default: false
+    config_param :checkpoint_interval, :integer, default: 30
+    # This is used to prune older entries that will be written to the sqlite db, the most recent entries(aka the hot LRU cache) will always be in the db.
+    config_param :checkpoint_ttl, :integer, default: 3000
```
dhananjaysathe: 3000 is a tad arbitrary as a default value imho

lib/fluent/plugin/checkpoint_metadata_cache.rb
```diff
@@ -0,0 +1,111 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+require "sqlite3"
+
+module KubernetesMetadata
+  module CheckPointCache
+    def start_checkpoint
+      while @checkpoint_thread_running
```
dhananjaysathe: Set and compare and actually use the mutex here in line with our discussion; return if already set

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -253,13 +269,39 @@ def log.trace?
 
     end
 
+    def shutdown
+      super
+      if @checkpoint_enabled
+        write_cache_to_file
+        @checkpoint_thread_mutex.synchronize do
```
dhananjaysathe: Fix this use of mutex too as discussed; In line with the comment above of the mutex and variable actually being used to do this

lib/fluent/plugin/filter_kubernetes_metadata.rb
HackToHell: @jcantrill Will this not be executed only once by default? (please ignore my current flawed implementation), but I don't get the requirement for making sure that `thread_running` is set to true only once, I think I am missing the user case for this.

lib/fluent/plugin/checkpoint_metadata_cache.rb
```diff
@@ -0,0 +1,111 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+require "sqlite3"
+
+module KubernetesMetadata
+  module CheckPointCache
```
jcantrill: Why are we calling this `CheckPointCache`instead of something more meaningful IMO, like `PersistentCache` or `PersistentLruCache`.  This is what this feature adds correct?  A cache that survives restarts of fluentd and rolls off old entries that have timed out.

lib/fluent/plugin/filter_kubernetes_metadata.rb
jcantrill: I defer to your findings here as I was not the original author of this plugin and was making a casual observation.  I would expect initialization only to occur again if say the process was restarted.  Alternatively if fluent properly supported multi-threading this could additionally be an issue

HackToHell: From what I have gone through other plugins, they all make the same assumption of just one thread having this. So I will remove the mutex in the code. As for the initialization happening multiple times, the init function is idempotent.

lib/fluent/plugin/checkpoint_metadata_cache.rb
HackToHell: Agreed, changed the name to something more meaningful.

jcantrill:
@HackToHell I realize we are slow to getting back to this PR but do you have interest in continuing to pursue this change?  If not, please close the PR and issue.

jcantrill:
Closing per https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/pull/157#issuecomment-572577457  Please re-open if needed

PR #169, opened 2019-04-17
WIP: remove namespace_id, container_meta:  image and image_id, labels, hos…

szilvesztersimicza:
remove namespace_id, container_meta:  image and image_id, labels, host, master_url fields from kubernetes metadata

### DIFF
--- a/lib/fluent/plugin/kubernetes_metadata_common.rb
+++ b/lib/fluent/plugin/kubernetes_metadata_common.rb
@@ -39,7 +39,6 @@ module KubernetesMetadata
         self.de_dot!(annotations)
       end
       kubernetes_metadata = {
-        'namespace_id' => namespace_object['metadata']['uid'],
         'creation_timestamp' => namespace_object['metadata']['creationTimestamp']
       }
       kubernetes_metadata['namespace_labels'] = labels unless labels.empty?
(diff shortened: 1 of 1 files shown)


jcantrill:
@szilvesztersimicza Can you please provide justification for this PR?  You are removing several of the fields which primarily why this plugin exists in the first place.  Minimally, you will need to provide configuration to exclude them, but these changes will break this plugins use with Openshift.

richm:
If you want to reduce the metadata, you should write a record_transformer or record_modifier filter to remove those fields you do not want

szilvesztersimicza:
For my work I do not need these fields. And I can use the record_transformer or filter_record_modifier plugin to remove these fields, but it is more cpu effective to not put these fields into the metadata at all.

But i do not need this PR.  I opened it, to use the CI, but it is outdated (CircleCI 1.0 -> 2.0), and the config process was unsuccessful.
Can Anyone close this PR?
Thanks!

richm:
> For my work I do not need these fields. And I can use the record_transformer or filter_record_modifier plugin to remove these fields, but it is more cpu effective to not put these fields into the metadata at all.

Then you could submit another PR which adds a couple of config parameters: a whitelist of fields to add (_only_ add these fields), and a blacklist (do not add these fields, but add all of the others); then change the code to use these new parameters when adding metadata fields. 

> 
> But i do not need this PR. I opened it, to use the CI, but it is outdated (CircleCI 1.0 -> 2.0), and the config process was unsuccessful.
> Can Anyone close this PR?
> Thanks!

done

PR #171, opened 2019-05-06
Reduce metadata 

szilvesztersimicza:
Our intention is to introduce some config options to be able to skip some fields in order to save significant storage capacity and hopefully some CPU cycles as well. Using default values the original behavior won't be changed.

We haven't found any alternative plugin to use instead this one. We know that additional plugins can be put at the end of the processing chain to filter out unnecessary fields, but we feel that it would be optimal (from CPU perspective) to not to collect them at a first place.

This feature is using 4 config parameter to skip some field of the metadata.  If these parameters (as the default value) are set to false, then the plugin works normally. But these parameters can switch independently the labels, container_metadata, master_url and namespace_metadata.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,10 @@ when true (default: `true`)
 * `ssl_partial_chain` - if `ca_file` is for an intermediate CA, or otherwise we do not have the root CA and want
   to trust the intermediate CA certs we do have, set this to `true` - this corresponds to
   the `openssl s_client -partial_chain` flag and `X509_V_FLAG_PARTIAL_CHAIN` (default: `false`)
-
+* `skip_labels` - Skip all label fields from the metadata.
+* `skip_container_metadata` - Skip some of the container data of the metadata. The metadata will not contain the container_image and container_image_id fields.
+* `skip_master_url` - Skip the master_url field from the metadata.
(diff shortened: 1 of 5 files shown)


jcantrill:
I am missing the point of this PR.  If you don't want metadata then consider not using the plugin. The value of the plugin is in the name.

szilvesztersimicza:
> I am missing the point of this PR. If you don't want metadata then consider not using the plugin. The value of the plugin is in the name.

I want some of the metadata, but not all of it. So this would be a configurable feature to skip some of the fields.

szilvesztersimicza:
Feature development is done. 
Please review it!

jcantrill:
Please provide tests to support your changes.  Additionally consider reimplementing by providing an alternate strategy to cache and fetch meta that doesn't require interlacing  configuration and "if" statements

szilvesztersimicza:
We are waiting for merge the circleci 2.0 request to the fabric8io:master. Tests are green.

szilvesztersimicza:
@jcantrill The circleCi 2.0 is configured and merged. A new testcase has been added and all testcases are green. Please merge it!

szilvesztersimicza:
@jcantrill is it ok for you?

szilvesztersimicza:
Thx for the merge!
When will be created the new version? So I can use "gem install" to install this feature.

jcantrill:
https://rubygems.org/gems/fluent-plugin-kubernetes_metadata_filter/versions/2.2.0

PR #173, opened 2019-05-15
add circleci 2.0

szilvesztersimicza:
The CircleCI 1.0 is no longer available. So this is a migration PR to upgrade to CircleCI 2.0.
https://circleci.com/sunset1-0/


.circleci/config.yml
```diff
@@ -0,0 +1,114 @@
+version: 2.0
+
+install: &install
```
richm: what does the `&install` here mean? I'm not familiar with this syntax

.circleci/config.yml
```diff
@@ -0,0 +1,114 @@
...
+      - run: *missingdeps
+      - checkout
+      - run: *install
+      - run: *test
+      - store_test_results:
+            path: coverage
+      - store_artifacts:
+            path: coverage
+
+  "ruby-2.3.8":
```
richm: Do we really need to test ruby 2.3 and earlier?  What major platforms still require ruby 2.3 and earlier?

.circleci/config.yml
szilvesztersimicza: It is like a reference or label, so defined once and used many times.

.circleci/config.yml
szilvesztersimicza: It was in the  circleci 1.0 config file, just migrated them to the 2.0 config.
We will be more than happy throw old ruby versions away.

szilvesztersimicza:
I have done some updates. Please Review it and if its correct please merge it! 
Thank You!

Issue #170, opened 2019-04-24
Improve memory usage


janario:
We've had a scenario where there were a lot of pod(started from cron) in `Pending` state. Around 9k

I know it was an internal scenario problem, but because of that we notice that fluentd started to crash (oomkiller)

Trying to understand the scenario and removing parts of our fluentd configuration we notice that filter_metadata was the problem and because we had too many pods

It would be good to have a low memory consumption

richm:
> We've had a scenario where there were a lot of pod(started from cron) in `Pending` state. Around 9k
> 
> I know it was an internal scenario problem, but because of that we notice that fluentd started to crash (oomkiller)
> 
> Trying to understand the scenario and removing parts of our fluentd configuration we notice that filter_metadata was the problem and because we had too many pods

How do you know it was the fluent-plugin-kubernetes_metadata_filter which was the problem?  Was the OOM kill stacktrace in this plugin code?

> 
> It would be good to have a low memory consumption

Have you tried adjusting `cache_size` and `cache_ttl`?

janario:
I started to remove piece by piece from my fluentd.conf

Basically my config is:

```
<source> for apps
<source> for kube-system
<source> for one file not managed by kubernetes

<filter kubemetadata for apps and kube-system
<match to send to cloudwatch logs

```

removing the sources it kept crashing, then I removed just the filter and it stopped to crash

I didn't try with the `cache_size` options


---
my oom-killer logs
```
Apr 24 11:51:07 ip-10-0-74-72 kernel: filter_kuberne* invoked oom-killer: gfp_mask=0x14000c0(GFP_KERNEL), nodemask=(null),  order=0, oom_score_adj=975
Apr 24 11:51:07 ip-10-0-74-72 kernel: filter_kuberne* cpuset=4dd475111b9c266dadd3132047c2baba4f7afe2ec9e7d895d0efe76f9806d0cd mems_allowed=0
Apr 24 11:51:07 ip-10-0-74-72 kernel: CPU: 1 PID: 28750 Comm: filter_kuberne* Not tainted 4.14.97-90.72.amzn2.x86_64 #1
Apr 24 11:51:07 ip-10-0-74-72 kernel: Hardware name: Xen HVM domU, BIOS 4.2.amazon 08/24/2006
Apr 24 11:51:07 ip-10-0-74-72 kernel: Call Trace:
[...]

PR #177, opened 2019-05-29
Added selector for watch_pods

janario:
Avoid read all the cluster by selecting only the current worknode

I've tried this in a test cluster with 5k pods and 200mb of memory for fluentd

Without this it crashes with `OOMKilled`

Resources should not increase as the cluster scale but as the worknode supports it

---

I still have a problem while limiting `get_pods`, it seems to not take effect

### DIFF
--- a/lib/fluent/plugin/kubernetes_metadata_watch_pods.rb
+++ b/lib/fluent/plugin/kubernetes_metadata_watch_pods.rb
@@ -24,9 +24,15 @@ module KubernetesMetadata
     include ::KubernetesMetadata::Common
 
     def start_pod_watch
+      k8s_node_name = ENV['K8S_NODE_NAME']
       begin
-        resource_version = @client.get_pods.resourceVersion
-        watcher          = @client.watch_pods(resource_version)
+        field_selector   = k8s_node_name ? "spec.nodeName=#{k8s_node_name}" : nil
+
+        # FIXME: limit seems to not take effect
+        # resource_version = @client.get_pods(limit: 1).resourceVersion
+        # watcher          = @client.watch_pods(resource_version: resource_version, field_selector: field_selector)
+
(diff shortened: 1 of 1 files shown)


richm:
basically, fluentd should only be interested in the pods and namespaces on its node - that is, there is no reason for fluentd running on node A to watch all of the pods/namespaces running on node B, node C, etc.

richm:
we will need some sort of test for this too, and an update for the watch related docs (if any, haven't checked)

janario:
@richm ideas on how to create a test for it?

I have to mock the kubernetes api and apply the filter. Do you have some similar test? 

Also can you help me with get_pods with limit? 
(I am not a ruby guy :-) )

richm:
maybe the problem is that this plugin is using a *very* old version of kubeclient?  https://github.com/abonas/kubeclient/blob/v1.1.4/lib/kubeclient/common.rb#L174 does not have support for `limit` afaict

Maybe you could help us find out if we can use the latest kubeclient?  https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/blob/master/fluent-plugin-kubernetes_metadata_filter.gemspec#L24
probably lots of other dependencies that need updating

janario:
@richm sorry for the delay

In the end we decided to disable the `watch` of pods since it was consuming too much memory as the cluster grows

I'm not a ruby developer so it would take too long to me to finish this. Feel free to implements the test, updates and requirements or just leave it open

At least I could exemplify the whole discussion started from https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/issues/170

richm:
We have recently updated the kubeclient version - perhaps this version supports using `watch` with limit?

richm:
closing in favor of https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/pull/189

PR #182, opened 2019-07-18
Update kubeclient version

baileyjensen-wk:
## Title
Update kubeclient version

### Problem
The previous version does not have the same breadth of capabilities that the up-to-date version has.

### Solution
Change the kubeclient version being used to the most up to date version.
(I would also be open to using slightly older kubeclient versions if there is good reason for doing so, but the current version is extremely out of date.)


richm:
closed in favor of https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/pull/185

PR #189, opened 2019-09-24
Improve Kubernetes pod lookups

### LINKED ISSUE #187: kubernetes_metadata plugin initial pull and cache strategy issue
Hi,

Recently we noticed some intriguing behavior based on apiserver logs that raised some performance concern:

The `kubernetes_metadata` plugin at startup lists all pods **with no filter** which caused load pressure on etcd for large clusters. After startup (even though it looks like a watch has been set up), it queries the API for certain pods every minute.

Looking at the code, it looks like what the kubernetes_metadata plugin intends to do is:
1. Initially queries API to get all pods and the resource version. (The initial pull **has no filters**. Besides that, it **does not seem to add these results to cache**.)
    Related code: https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/blob/8a75c3d5811b6970ef7d11a42c43a74d3902b839/lib/fluent/plugin/kubernetes_metadata_watch_pods.rb#L28
2. Sets up a pod watch with that resource version and keeps updating the cache.
    Related code: https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/blob/8a75c3d5811b6970ef7d11a42c43a74d3902b839/lib/fluent/plugin/kubernetes_metadata_watch_pods.rb#L29
3. Only queries API again with pod name filter if the pod is not already in the cache. (In theory this should be rare, which contradicts what we saw from the log: The plugin queries the API every minute. Perhaps this is because we did not cache the results in step 1?)
    Related code: https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/blob/f1831d66baeee3540ca737e42b009d43c0513878/lib/fluent/plugin/kubernetes_metadata_cache_strategy.rb#L29

**Proposal**

1. In the initial "get_pods", filter by local nodes. (I noticed that in the README, it says this plugin also supports `multi-tenant`. I'm not sure if filtering by local nodes would affect that feature).
2. Cache the `get_pods` results from the initial poll. (Unless I read the code wrong, it does not seem to cache it).

**Our configuration**

```
    <filter reform.**>
      @type kubernetes_metadata
      # Interval in seconds to dump cache stats locally in the Fluentd log.
      stats_interval 300
      # TTL in seconds of each cached element.
      cache_ttl 30
      # Skip fetching unused metadata.
      skip_container_metadata true
      skip_master_url true
      skip_namespace_metadata true
    </filter>
```

**Impact**

Load pressure on etcd + Fluentd OOM crash loops for large clusters in the initial pull.

jkohen:
This PR contains the following related fixes to pod lookup performance:
1. Cache results from `get_pods`.
1. Restrict `get_pods` and `watch_pods` requests to the current host, if provided with the `K8S_NODE_NAME` environment variable.
1. Force lookup from API server cache by setting resourceVersion=0 in `get_pods` request.

Item #3 is only effective with https://github.com/abonas/kubeclient/pull/420, otherwise it's a harmless no-op (kubeclient will ignore the option).

Closes #187

### DIFF
--- a/lib/fluent/plugin/kubernetes_metadata_watch_pods.rb
+++ b/lib/fluent/plugin/kubernetes_metadata_watch_pods.rb
@@ -24,22 +24,24 @@ module KubernetesMetadata
     include ::KubernetesMetadata::Common
 
     def start_pod_watch
-      field_selector = ''
       if ENV['K8S_NODE_NAME']
         field_selector = 'spec.nodeName=' + ENV['K8S_NODE_NAME']
       end
       begin
-        pods = @client.get_pods(
-          resource_version: 0,  # Fetch from API server.
-          field_selector: field_selector)
+        options = {
+          resource_version: 0  # Fetch from API server.
+        }
+        if field_selector
+          options[:field_selector] = field_selector
+        end
+        pods = @client.get_pods(options)
         pods.each do |pod|
           cache_key = pod.metadata['uid']
           @cache[cache_key] = parse_pod_metadata(pod)
           @stats.bump(:pod_cache_host_updates)
         end
-        watcher = @client.watch_pods(
-          resource_version: pods.resourceVersion,
-          field_selector: field_selector)
+        options[:resource_version] = pods.resourceVersion
+        watcher = @client.watch_pods(options)
       rescue Exception => e
         message = "Exception encountered fetching metadata from Kubernetes API endpoint: #{e.message}"
         message += " (#{e.response})" if e.respond_to?(:response)
--- a/lib/fluent/plugin/kubernetes_metadata_watch_pods.rb
+++ b/lib/fluent/plugin/kubernetes_metadata_watch_pods.rb
@@ -25,8 +25,15 @@ module KubernetesMetadata
 
     def start_pod_watch
       begin
-        resource_version = @client.get_pods.resourceVersion
-        watcher          = @client.watch_pods(resource_version)
+        pods = @client.get_pods
+        pods.each do |pod|
+          cache_key = pod.metadata['uid']
+          if ENV['K8S_NODE_NAME'] == pod.spec['nodeName'] then
+            @cache[cache_key] = parse_pod_metadata(pod)
+            @stats.bump(:pod_cache_host_updates)
+          end
+        end
+        watcher = @client.watch_pods(pods.resourceVersion)
       rescue Exception => e
         message = "Exception encountered fetching metadata from Kubernetes API endpoint: #{e.message}"
         message += " (#{e.response})" if e.respond_to?(:response)
--- a/lib/fluent/plugin/kubernetes_metadata_watch_pods.rb
+++ b/lib/fluent/plugin/kubernetes_metadata_watch_pods.rb
@@ -29,7 +29,7 @@ module KubernetesMetadata
       end
       begin
         options = {
-          resource_version: 0  # Fetch from API server.
+          resource_version: '0'  # Fetch from API server.
         }
         if field_selector
           options[:field_selector] = field_selector
--- a/lib/fluent/plugin/kubernetes_metadata_watch_pods.rb
+++ b/lib/fluent/plugin/kubernetes_metadata_watch_pods.rb
@@ -54,9 +54,11 @@ module KubernetesMetadata
             if cached
               @cache[cache_key] = parse_pod_metadata(notice.object)
               @stats.bump(:pod_cache_watch_updates)
-            else
+            elsif ENV['K8S_NODE_NAME'] == notice.object['spec']['nodeName'] then
               @cache[cache_key] = parse_pod_metadata(notice.object)
               @stats.bump(:pod_cache_host_updates)
+            else
+              @stats.bump(:pod_cache_watch_misses)
             end
           when 'DELETED'
             # ignore and let age out for cases where pods
--- a/lib/fluent/plugin/kubernetes_metadata_watch_pods.rb
+++ b/lib/fluent/plugin/kubernetes_metadata_watch_pods.rb
@@ -29,13 +29,17 @@ module KubernetesMetadata
         field_selector = 'spec.nodeName=' + ENV['K8S_NODE_NAME']
       end
       begin
-        pods = @client.get_pods(field_selector: field_selector)
+        pods = @client.get_pods(
+          resource_version: 0,  # Fetch from API server.
+          field_selector: field_selector)
         pods.each do |pod|
           cache_key = pod.metadata['uid']
           @cache[cache_key] = parse_pod_metadata(pod)
           @stats.bump(:pod_cache_host_updates)
         end
-        watcher = @client.watch_pods(pods.resourceVersion, field_selector: field_selector)
+        watcher = @client.watch_pods(
+          resource_version: pods.resourceVersion,
+          field_selector: field_selector)
       rescue Exception => e
         message = "Exception encountered fetching metadata from Kubernetes API endpoint: #{e.message}"
         message += " (#{e.response})" if e.respond_to?(:response)
--- a/lib/fluent/plugin/kubernetes_metadata_watch_pods.rb
+++ b/lib/fluent/plugin/kubernetes_metadata_watch_pods.rb
@@ -24,15 +24,12 @@ module KubernetesMetadata
     include ::KubernetesMetadata::Common
 
     def start_pod_watch
-      if ENV['K8S_NODE_NAME']
-        field_selector = 'spec.nodeName=' + ENV['K8S_NODE_NAME']
-      end
       begin
         options = {
           resource_version: '0'  # Fetch from API server.
         }
-        if field_selector
-          options[:field_selector] = field_selector
+        if ENV['K8S_NODE_NAME']
+          options[:field_selector] = 'spec.nodeName=' + ENV['K8S_NODE_NAME']
         end
         pods = @client.get_pods(options)
         pods.each do |pod|
--- a/lib/fluent/plugin/kubernetes_metadata_watch_pods.rb
+++ b/lib/fluent/plugin/kubernetes_metadata_watch_pods.rb
@@ -24,16 +24,18 @@ module KubernetesMetadata
     include ::KubernetesMetadata::Common
 
     def start_pod_watch
+      field_selector = ''
+      if ENV['K8S_NODE_NAME']
+        field_selector = 'spec.nodeName=' + ENV['K8S_NODE_NAME']
+      end
       begin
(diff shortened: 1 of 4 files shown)


lib/fluent/plugin/kubernetes_metadata_watch_pods.rb
```diff
@@ -24,9 +24,24 @@ module WatchPods
     include ::KubernetesMetadata::Common
 
     def start_pod_watch
+      if ENV['K8S_NODE_NAME']
+        field_selector = 'spec.nodeName=' + ENV['K8S_NODE_NAME']
```
richm: is `field_selector` used anywhere except below?

lib/fluent/plugin/kubernetes_metadata_watch_pods.rb
```diff
@@ -24,9 +24,24 @@ module WatchPods
...
+      end
       begin
-        resource_version = @client.get_pods.resourceVersion
-        watcher          = @client.watch_pods(resource_version)
+        options = {
+          resource_version: '0'  # Fetch from API server.
+        }
+        if field_selector
+          options[:field_selector] = field_selector
```
richm: if this is the only place `field_selector` is used, then change this block to
```ruby
if ENV['K8S_NODE_NAME']
    options[:field_selector] = 'spec.nodeName=' + ENV['K8S_NODE_NAME']
fi
```

jkohen: Makes sense, thanks!

lib/fluent/plugin/kubernetes_metadata_watch_pods.rb
```diff
@@ -42,11 +54,9 @@ def start_pod_watch
             if cached
               @cache[cache_key] = parse_pod_metadata(notice.object)
               @stats.bump(:pod_cache_watch_updates)
-            elsif ENV['K8S_NODE_NAME'] == notice.object['spec']['nodeName'] then
+            else
               @cache[cache_key] = parse_pod_metadata(notice.object)
               @stats.bump(:pod_cache_host_updates)
-            else
-              @stats.bump(:pod_cache_watch_misses)
```
jcantrill: Can we guarantee we will NEVER have a cache miss?

lib/fluent/plugin/kubernetes_metadata_watch_pods.rb
```diff
@@ -25,8 +25,20 @@ module WatchPods
...
       begin
-        resource_version = @client.get_pods.resourceVersion
-        watcher          = @client.watch_pods(resource_version)
+        options = {
+          resource_version: '0'  # Fetch from API server.
+        }
+        if ENV['K8S_NODE_NAME']
+          options[:field_selector] = 'spec.nodeName=' + ENV['K8S_NODE_NAME']
+        end
+        pods = @client.get_pods(options)
```
jcantrill: My familiarity with the ruby client is limited but this is enough to restrict the list of pods to a given node?  I have not seen such syntax with any of the other clients or the REST api

lib/fluent/plugin/kubernetes_metadata_watch_pods.rb
jkohen: We can't, good catch. Originally I assumed that K8S_NODE_NAME was always set, but then I learned it isn't. I've restored the code and the test. Sorry about that!

lib/fluent/plugin/kubernetes_metadata_watch_pods.rb
jkohen: It is. E.g. https://github.com/kubernetes/client-go/issues/410
It is also recommended by the Google Kubernetes team. @wojtek-t @krzysied is there any reference you can share?

lib/fluent/plugin/kubernetes_metadata_watch_pods.rb
jkohen: On second thought, I think this code is redundant indeed. If K8S_NODE_NAME is set, there should be do misses due to filtering in the API. If it's not set, then the elsif block won't run. Your call whether to leave it or not.

lib/fluent/plugin/kubernetes_metadata_watch_pods.rb
```diff
@@ -25,8 +25,20 @@ module WatchPods
 
     def start_pod_watch
       begin
-        resource_version = @client.get_pods.resourceVersion
-        watcher          = @client.watch_pods(resource_version)
+        options = {
+          resource_version: '0'  # Fetch from API server.
```
richm: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.12/#list-pod-v1-core
```
resourceVersion | ... - if it's 0,  then we simply return what we currently have in cache, no guarantee;
```
so it is fast, but "unsafe"
I don't think we want to hardcode this value to `'0'` - it should at least be settable by the user, in case there are users who must have up-to-date data, or who want to use whatever is the cluster policy e.g.
```
When specified for list: - if unset, then the result is returned from remote storage based on quorum-read flag;
```
I'm assuming `get_pods` is doing a `list` command.  So I think the user should have the option of either setting it to `'0'`, or omitting `resource_version` entirely from the query.

lib/fluent/plugin/kubernetes_metadata_watch_pods.rb
richm: There's https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.12/#list-pod-v1-core
```
fieldSelector | A selector to restrict the list of returned objects by their fields. Defaults to everything.
```

lib/fluent/plugin/kubernetes_metadata_watch_pods.rb
jkohen: @wojtek-t can you please comment on the implications of using the cache, especially on data freshness and how a user would choose whether to set `resourceVersion` to 0 or leave it unset?

@richm how should I make this configurable? A plug-in option?

lib/fluent/plugin/kubernetes_metadata_watch_pods.rb
wojtek-t: fieldSelector is widely used in k8s codebase.
As an example, kubelet is doing exactly that (list + watch of its pods) and is using fieldSelector
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kubelet.go#L456
[obviously not in ruby]

lib/fluent/plugin/kubernetes_metadata_watch_pods.rb
wojtek-t: We don't provide explicit guarantee - that's correct.
But in practice, there is guarantee - as an example, we have 5s pod startup SLO: https://github.com/kubernetes/community/blob/master/sig-scalability/slos/pod_startup_latency.md
Why is that related? Basically starting a pod requires a couple rounds of watch, which is served exactly from the same cache.
So practice, in order to meet this SLO, it has to be in the order of tens-hundreds milliseconds, and that is what is happening.

Also, *all* Kubernetes component are working in that mode (e.g. it's build in the the reflector and informer frameworks). Kubelet, scheduler, controllers being part of controller-manager, ... all operate in that mode:
- list from cache: with RV=0
- start watching from that point
- reconcile state based on that

So you're not introducing anything new - I would even say "you're very late in this game".
I don't consider it unsafe in any means.

richm: > I would even say "you're very late in this game".

So all "officially supported" kubernetes api watch clients use `resourceVersion: 0`?  @bparees do you know anything about this?  Is this what the operator sdk uses to set up watches?  Who would we talk to in the operator team about this?  I don't think it is prudent to be in the situation where we are the only openshift component using watches in this way.  I did a quick scan through the cluster-logging-operator vendored code and it isn't clear to me that it uses watches with `resourceversion: 0`

jkohen: To be more specific, the LIST request uses RV=0. The WATCH uses the RV in the response from LIST. This client already has the intended behavior for WATCH. The new behavior is to set RV=0 in the LIST request.

bparees: @richm it's safe elsewhere because we are always "list/watching" (list what's there, and then watch for changes).  If you're just listing (and thus not going to then watch for/consume changes that happened after the version you got back in the list) then i would agree that it's unsafe.  i'm not sure which pattern is being used here?

bparees: another note is that there's basically no difference between "we asked for the latest and got it, but then 2ms later the resource was updated" and "we asked for the cached version and got something 2ms out of date".   W/o more context on exactly what we're doing w/ this value it's hard for me to comment intelligently, but i'd be wary of any thought process which is "we have to get the latest value when we invoke the get because we can't tolerate getting back a slightly stale value".

wojtek-t: Correct - I was talking about exactly this case which is "list + watch". I agree that if you need strong consistency or just doing LIST without watching after it, it may not always be safe.

richm:
ok - lgtm - please squash into a single commit - jcantrill - any objections?

jkohen:
@richm can you squash on merge? I believe that's an option in github. See #3 under https://help.github.com/en/articles/merging-a-pull-request#merging-a-pull-request-on-github

richm:
> @richm can you squash on merge?

Looks like it - cool - never knew about that option.  @jcantrill @bparees any objections?

bparees:
>  Looks like it - cool - never knew about that option. @jcantrill @bparees any objections?

nope, it's fine..just not usually an option because we bot-merge elsewhere.  but if this repo uses button-merging, sure.

jkohen:
@richm we really appreciate the super quick review. When do you release this repo? I'll want to update our package to use that version of the GEM.

richm:
I think a change of this scope warrants a 2.4.0 release.  Can cut a new release in a few hours once everyone approves.  I do not know the maintainers of the `kubeclient`

jkohen:
Whose approval are we waiting for? I got ahold of the `kubeclient` maintainer.

richm:
> Whose approval are we waiting for? I got ahold of the `kubeclient` maintainer in the corresponding PR.

@jcantrill is the primary maintainer of this code - I want to make sure he is ok with it

jcantrill:
https://rubygems.org/gems/fluent-plugin-kubernetes_metadata_filter/versions/2.4.0

PR #193, opened 2019-09-30
get rid of dependency on activesupport

richm:
Some transitive dependency on some newer project has crept in
via activesupport.  We only use activesupport for a couple of
trivial methods, so just implement those methods in our code.
Due to needing to support ruby 2.0.0, we have to lock down the
versions of several of the components we use.


richm:
@jcantrill this is now blocking ocp 3.11 upstream builds/ci

richm:
remove /hold - it is working - please review

PR #198, opened 2019-11-13
Adds VERIFY_SSL env config support

protomonk:
There are several issue reported on fluentd repo regarding issues with internal Kubernetes certificate verification. Something like this will help fix the things without the need to rebuild and maintain copy of fluentd-daemonset docker images.

### DIFF
--- a/lib/fluent/plugin/filter_kubernetes_metadata.rb
+++ b/lib/fluent/plugin/filter_kubernetes_metadata.rb
@@ -204,6 +204,11 @@ module Fluent::Plugin
         end
       end
 
+      # Use verify_ssl configuration from env variable
+      if ENV['VERIFY_SSL'].present?
+      	@verify_ssl = ENV['VERIFY_SSL'] == "true" ? true : false
+      end
+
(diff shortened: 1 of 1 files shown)


lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -204,6 +204,11 @@ def log.trace?
         end
       end
 
+      # Use verify_ssl configuration from env variable
+      if ENV['VERIFY_SSL'].present?
+      	@verify_ssl = ENV['VERIFY_SSL'] == "true" ? true : false
```
richm:
```ruby
@verify_ssl = ENV['VERIFY_SSL'] == "true"
```
the comparison already returns a boolean value so there is no need to use the trinary operator.

Also - I would suggest perhaps using a different environment variable name so as not to conflict with other possible uses - other applications might use VERIFY_SSL - perhaps K8S_META_VERIFY_SSL?

jcantrill:
Closing stale with no recent response

Issue #194, opened 2019-10-02
Fluentd Crash


emillg:
Fluentd crashed with log:
```
#<Thread:0x00007fa17a7d3778@/usr/local/bundle/gems/fluent-plugin-kubernetes_metadata_filter-2.4.0/lib/fluent/plugin/filter_kubernetes_metadata.rb:265 run> terminated with exception (report_on_exception is true):
/usr/local/bundle/gems/http-4.1.1/lib/http/response/parser.rb:14:in `<<': error reading from socket: Could not parse data entirely (1 != 0) (HTTP::ConnectionError)
	from /usr/local/bundle/gems/http-4.1.1/lib/http/response/parser.rb:14:in `add'
	from /usr/local/bundle/gems/http-4.1.1/lib/http/connection.rb:214:in `read_more'
	from /usr/local/bundle/gems/http-4.1.1/lib/http/connection.rb:92:in `readpartial'
	from /usr/local/bundle/gems/http-4.1.1/lib/http/response/body.rb:30:in `readpartial'
	from /usr/local/bundle/gems/http-4.1.1/lib/http/response/body.rb:36:in `each'
	from /usr/local/bundle/gems/kubeclient-4.5.0/lib/kubeclient/watch_stream.rb:25:in `each'
	from /usr/local/bundle/gems/fluent-plugin-kubernetes_metadata_filter-2.4.0/lib/fluent/plugin/kubernetes_metadata_watch_namespaces.rb:36:in `start_namespace_watch'
	from /usr/local/bundle/gems/fluent-plugin-kubernetes_metadata_filter-2.4.0/lib/fluent/plugin/filter_kubernetes_metadata.rb:265:in `block in configure'
/usr/local/bundle/gems/http-4.1.1/lib/http/response/parser.rb:14:in `<<': Could not parse data entirely (1 != 0) (HTTP::Parser::Error)
	from /usr/local/bundle/gems/http-4.1.1/lib/http/response/parser.rb:14:in `add'
	from /usr/local/bundle/gems/http-4.1.1/lib/http/connection.rb:214:in `read_more'
	from /usr/local/bundle/gems/http-4.1.1/lib/http/connection.rb:92:in `readpartial'
	from /usr/local/bundle/gems/http-4.1.1/lib/http/response/body.rb:30:in `readpartial'
	from /usr/local/bundle/gems/http-4.1.1/lib/http/response/body.rb:36:in `each'
[...]

PR #204, opened 2020-01-03
Add retry logic in pod and namespace watches when Kubernetes API connection gets closed.

qingling128:
Fixes https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/issues/194.

### DIFF
--- a/lib/fluent/plugin/filter_kubernetes_metadata.rb
+++ b/lib/fluent/plugin/filter_kubernetes_metadata.rb
@@ -291,6 +291,7 @@ module Fluent::Plugin
                 pod_watcher ||= get_pods_and_start_watcher
                 process_pod_watcher_notices(pod_watcher)
               rescue Exception => e
+                @stats.bump(:pod_watch_failures)
                 if Thread.current[:pod_watch_retry_count] <= @watch_retry_max_times
                   # Instead of raising exceptions and crashing Fluentd, swallow
                   # the exception and reset the watcher.
@@ -333,6 +334,7 @@ module Fluent::Plugin
                 namespace_watcher ||= get_namespaces_and_start_watcher
                 process_namespace_watcher_notices(namespace_watcher)
               rescue Exception => e
+                @stats.bump(:namespace_watch_failures)
                 if Thread.current[:namespace_watch_retry_count] <= @watch_retry_max_times
                   # Instead of raising exceptions and crashing Fluentd, swallow
                   # the exception and reset the watcher.
--- a/lib/fluent/plugin/filter_kubernetes_metadata.rb
+++ b/lib/fluent/plugin/filter_kubernetes_metadata.rb
@@ -85,7 +85,11 @@ module Fluent::Plugin
     config_param :skip_master_url, :bool, default: false
     config_param :skip_namespace_metadata, :bool, default: false
     # The time interval in seconds for retry backoffs when watch connections fail.
-    config_param :watch_retry_interval, :bool, default: 10
+    config_param :watch_retry_interval, :bool, default: 1
+    # The base number of exponential backoff for retries.
+    config_param :watch_retry_exponential_backoff_base, :bool, default: 2
+    # The maximum number of times to retry pod and namespace watches.
+    config_param :watch_retry_max_times, :bool, default: 10
 
     def fetch_pod_metadata(namespace_name, pod_name)
       log.trace("fetching pod metadata: #{namespace_name}/#{pod_name}") if log.trace?
@@ -274,15 +278,40 @@ module Fluent::Plugin
             # Any failures / exceptions in the initial setup should raise
             # Fluent:ConfigError, so that users can inspect potential errors in
             # the configuration.
-            watcher = start_pod_watch
+            pod_watcher = start_pod_watch
+            Thread.current[:pod_watch_retry_backoff_interval] = @watch_retry_interval
+            Thread.current[:pod_watch_retry_count] = 0
 
             # Any failures / exceptions in the followup watcher notice
             # processing will be swallowed and retried. These failures /
             # exceptions could be caused by Kubernetes API being temporarily
             # down. We assume the configuration is correct at this point.
             while thread_current_running?
-              process_pod_watcher_notices_with_retries(
-                reset_pod_watcher_if_needed(watcher))
+              begin
+                pod_watcher ||= get_pods_and_start_watcher
+                process_pod_watcher_notices(pod_watcher)
+              rescue Exception => e
+                if Thread.current[:pod_watch_retry_count] <= @watch_retry_max_times
+                  # Instead of raising exceptions and crashing Fluentd, swallow
+                  # the exception and reset the watcher.
+                  log.info(
+                    "Exception encountered parsing pod watch event. The " \
+                    "connection might have been closed. Sleeping for " \
+                    "#{Thread.current[:pod_watch_retry_backoff_interval]} " \
+                    "seconds and resetting the pod watcher.", e)
+                  sleep(Thread.current[:pod_watch_retry_backoff_interval])
+                  Thread.current[:pod_watch_retry_count] += 1
+                  Thread.current[:pod_watch_retry_backoff_interval] *= @watch_retry_exponential_backoff_base
+                  pod_watcher = nil
+                else
+                  # Since retries failed for many times, log as errors instead
+                  # of info and raise exceptions and trigger Fluentd to restart.
+                  log.error(
+                    "Exception encountered parsing pod watch event. The " \
+                    "connection might have been closed. Retried " \
+                    "#{@watch_retry_max_times} times yet still failing.", e)
+                end
+              end
             end
           end
           pod_thread.abort_on_exception = true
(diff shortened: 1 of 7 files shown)


lib/fluent/plugin/kubernetes_metadata_watch_pods.rb
```diff
@@ -24,28 +24,44 @@ module WatchPods
...
+      log.debug(message)
+
+      raise Fluent::ConfigError, message
+    end
 
-        raise Fluent::ConfigError, message
+    # List all pods, record the resourceVersion and return a watcher starting
+    # from that resourceVersion.
+    def get_pods_and_start_watcher
+      options = {
```
jcantrill: I realize we are moving the code but why would we start from resourceVersion 0 instead of applying the same pattern that is used by getting namespaces:

1. list pods
1. get resourceversion
1. watch pods from resourceversion

Does 'watching from version 0' get every add/update/delete that exists instead of 'watch from now'?

fluent-plugin-kubernetes_metadata_filter.gemspec
```diff
@@ -4,7 +4,7 @@ $LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
 
 Gem::Specification.new do |gem|
   gem.name          = "fluent-plugin-kubernetes_metadata_filter"
-  gem.version       = "2.4.1"
+  gem.version       = "2.4.2.lingshi"
```
jcantrill: revert me

lib/fluent/plugin/kubernetes_metadata_watch_pods.rb
qingling128: @jcantrill - The current implementation seems to follow your suggestion:

1. list pods (line 47)
2. get resourceversion (line 53)
3. watch pods from resourceversion (line 54)

Only the initial `get_pods` is from `resourceVersion=0`, which was recommended by the Kubernetes team because that forced it to use apiserver's local cache, which is faster.

fluent-plugin-kubernetes_metadata_filter.gemspec
qingling128: Reverted.

lib/fluent/plugin/kubernetes_metadata_watch_pods.rb
```diff
@@ -24,28 +24,44 @@ module WatchPods
...
+      watcher = @client.watch_pods(options)
+      Thread.current[:watcher_alive] = true
+      watcher
+    end
 
+    # Given a watcher, process the notices and handle retries by resetting the
+    # watcher if needed.
+    def process_pod_watcher_notices_with_retries(watcher)
+      unless @watcher_alive
+        watcher = get_pods_and_start_watcher
```
jkohen: Nit, could you get the same behavior by storing the watcher in a `@watcher` variable?

lib/fluent/plugin/kubernetes_metadata_watch_pods.rb
```diff
@@ -70,6 +86,14 @@ def start_pod_watch
             @stats.bump(:pod_cache_watch_ignored)
         end
       end
+    rescue Exception => e
+      # Instead of raising exceptions and crashing Fluentd, swallow the
+      # exception and reset the watcher.
+      log.info "Exception encountered parsing pod watch event. " \
+               "The connection might have been closed. " \
+               "Resetting the pod watcher.", e
+      Thread.current[:watcher_alive] = false
```
jkohen: Why is this written through `Thread.current[:watcher_alive]`, but read with `@watcher_alive` above?

qingling128:
Pushed a commit to fix CircleCI as it's using Debian buster instead of Debian Jessie now. This eliminated the error (the error also exists for the master branch) below:

```
The following packages have unmet dependencies:
 cmake : Depends: libcurl3 (>= 7.16.2) but it is not going to be installed
E: Unable to correct problems, you have held broken packages.
```

Now it's failing for a different error:

```
#!/bin/bash -eo pipefail
bundle install --path vendor/bundle
Fetching gem metadata from https://rubygems.org/.........
Fetching gem metadata from https://rubygems.org/.
Resolving dependencies...
Bundler could not find compatible versions for gem "bundler":
  In Gemfile:
    bundler (~> 2.0.2)

  Current Bundler version:
    bundler (1.17.3)
This Gemfile requires a different version of Bundler.
Perhaps you need to update Bundler by running `gem install bundler`?

Could not find gem 'bundler (~> 2.0.2)' in any of the relevant sources:
  the local ruby installation

Exited with code exit status 6
```

lib/fluent/plugin/kubernetes_metadata_watch_pods.rb
```diff
@@ -24,28 +24,44 @@ module WatchPods
...
+        @stats.bump(:pod_cache_host_updates)
+      end
+      options[:resource_version] = pods.resourceVersion
+      watcher = @client.watch_pods(options)
+      Thread.current[:watcher_alive] = true
+      watcher
+    end
 
+    # Given a watcher, process the notices and handle retries by resetting the
+    # watcher if needed.
```
jkohen: I was confused by the API at first. It seems weird that this method takes a watcher, but can also instantiate its own. It would be clearer if it only did one of those things. Do you think it can be refactored?

qingling128: Sure thing. Refactored.

lib/fluent/plugin/kubernetes_metadata_watch_pods.rb
qingling128: `@watcher` is not thread safe I think. Something like `Thread.current[:watcher]` should work. I'll address this when adding unit tests.

lib/fluent/plugin/kubernetes_metadata_watch_pods.rb
qingling128: Fixed.

qingling128:
Pushed a commit to fix the `bundler` issue. Will add some unit tests and fix the remaining tests.

lib/fluent/plugin/kubernetes_metadata_watch_pods.rb
jcantrill: @qingling128 I defer to what you were advised but this is different from the namespace watch:
```    def get_namespaces_and_start_watcher
      resource_version = @client.get_namespaces.resourceVersion
      watcher = @client.watch_namespaces(resource_version)
      Thread.current[:namespace_watcher_alive] = true
      watcher
    end```

Comment (jkohen) [reply lib/fluent/plugin/kubernetes_metadata_watch_pods.rb]: That's probably because I didn't touch the namespaces file when I added the rv=0 logic to the pod watcher, as we hadn't noticed problems with it. It makes sense to do this in both since we're changing the code substantially.

File: lib/fluent/plugin/kubernetes_metadata_watch_pods.rb (hunk 5 above)
Comment (jkohen) [reply lib/fluent/plugin/kubernetes_metadata_watch_pods.rb]: Thanks, it's clearer now. Unless I'm missing something, I think it could benefit from being even simpler. E.g.:
```ruby
wacher = nil
while thread_current_running?
  watcher ||= get_namespaces_and_start_watcher
  begin
    process_namespace_watcher_notices_with_retries(watcher)
  rescue Exception => e
    log.info "Exception encountered parsing namespace watch event. " \
             "The connection might have been closed. " \
             "Resetting the namespace watcher.", e
    watcher = nil
    sleep(@watch_retry_interval)
  end
end
```

lib/fluent/plugin/kubernetes_metadata_watch_pods.rb
```diff
@@ -70,6 +92,14 @@ def start_pod_watch
             @stats.bump(:pod_cache_watch_ignored)
         end
       end
+    rescue Exception => e
+      # Instead of raising exceptions and crashing Fluentd, swallow the
+      # exception and reset the watcher.
+      log.info "Exception encountered parsing pod watch event. " \
+               "The connection might have been closed. " \
+               "Resetting the pod watcher.", e
+      Thread.current[:pod_watcher_alive] = false
```
bmoyles0117: @jcantrill, Ling and I had a discussion offline, here's our takeaways:

Adding the two metrics below would allow us to approximate downtime for API servers.
- retry_sleep_total metric to sum up time that we're sleeping
- retry_attempts metric to sum up the number of reconnects

We could consider allowing the original exception to be thrown if the time slept exceeds a critical threshold.

jcantrill: > @jcantrill, Ling and I had a discussion offline, here's our takeaways:
> 
> Adding the two metrics below would allow us to approximate downtime for API servers.
> 
>     * retry_sleep_total metric to sum up time that we're sleeping
> 
>     * retry_attempts metric to sum up the number of reconnects

WFM.  The metrics that exist are not exposed to anything really consumable since it was part of the initial caching impl to get some details about what was going on.  Future task is likely to make the prometheus consumable.

> 
> 
> We could consider allowing the original exception to be thrown if the time slept exceeds a critical threshold.

In the past where I have utilized an exponential backoff, I have simply started the backoff loop over.  I don't know if there is value in throwing the exception but there is at least value in logging it every so many times we see it.

jcantrill:
> Pushed a commit to fix CircleCI as it's using Debian buster instead of Debian Jessie now. This eliminated the error (the error also exists for the master branch) below:
> 
> ```
> The following packages have unmet dependencies:
>  cmake : Depends: libcurl3 (>= 7.16.2) but it is not going to be installed
> E: Unable to correct problems, you have held broken packages.
> ```
> 
> Now it's failing for a different error:
> 
> ```
> #!/bin/bash -eo pipefail
> bundle install --path vendor/bundle
> Fetching gem metadata from https://rubygems.org/.........
> Fetching gem metadata from https://rubygems.org/.
> Resolving dependencies...
> Bundler could not find compatible versions for gem "bundler":
>   In Gemfile:
>     bundler (~> 2.0.2)
> 
>   Current Bundler version:
>     bundler (1.17.3)
> This Gemfile requires a different version of Bundler.
> Perhaps you need to update Bundler by running `gem install bundler`?
> 
> Could not find gem 'bundler (~> 2.0.2)' in any of the relevant sources:
>   the local ruby installation
> 
> Exited with code exit status 6
> ```
Please submit  a separte PR to fix up the tests if you have a change to resolve that issue.

qingling128:
Split the CircleCI fix to https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/pull/206.

qingling128:
Split the namespace lookup improvements into https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/pull/209. Will rebase this one once the other one is merged first.

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -80,6 +84,8 @@ class KubernetesMetadataFilter < Fluent::Plugin::Filter
     config_param :skip_container_metadata, :bool, default: false
     config_param :skip_master_url, :bool, default: false
     config_param :skip_namespace_metadata, :bool, default: false
+    # The time interval in seconds for retry backoffs when watch connections fail.
+    config_param :watch_retry_interval, :bool, default: 10
```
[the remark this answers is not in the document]
qingling128: Changed to 1 and added backoff.

lib/fluent/plugin/kubernetes_metadata_watch_namespaces.rb
```diff
@@ -54,7 +77,14 @@ def start_namespace_watch
...
         end
       end
+    rescue Exception => e
+      # Instead of raising exceptions and crashing Fluentd, swallow the
+      # exception and reset the watcher.
+      log.info "Exception encountered parsing namespace watch event. " \
+               "The connection might have been closed. " \
+               "Resetting the namespace watcher.", e
+      Thread.current[:namespace_watcher_alive] = false
+      sleep(@watch_retry_interval)
```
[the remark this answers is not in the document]
qingling128: Changed to 1 and added backoff.

lib/fluent/plugin/kubernetes_metadata_watch_pods.rb
qingling128: Refactored to move exception handling to the main filter file. Also added retry_backoff and retry_max_times as suggested in the other thread.

lib/fluent/plugin/kubernetes_metadata_watch_pods.rb
qingling128: Added two metrics: pod_watch_failures and namespace_watch_failures

Added retry_backoff and retry_max_times. I changed it to log an error for every 10 retries (roughly 17 minutes) and raise an exception to let fluentd restart. This way it something is completely wrong, a Fluentd restart could hopefully get it fixed.

qingling128:
Addressed feedback. This is ready for review now.

jcantrill:
@jkohen lgtm. any reason not to merge?

jkohen:
> @jkohen lgtm. any reason not to merge?

Lgtm. The new code structure is much easier to follow. Thanks, @qingling128 !

qingling128:
@jcantrill You are welcome. Do we know when the next release would be?

jcantrill:
https://rubygems.org/gems/fluent-plugin-kubernetes_metadata_filter/versions/2.4.2

qingling128:
Thanks!

PR #207, opened 2020-01-11
loop watch and performance improvement

rockb1017:
This is effort by cisco developers (Stephen & Parjinder).
The impact of these changes are;
* when pod and namespace metadata is updated on kubernetes (ie, namespace being annotated), it gets picked up right away. (I am not sure why version without this patch doesn't pick up instantly) 
* it uses much less CPU.


jcantrill:
@rockb1017 please see #204.  There's probably room for consolidation

rockb1017:
okay. I can test with #204 and see if we need this PR.

jcantrill:
@rockb1017 is there more to be done here or can we close this PR?

rockb1017:
nope. thank you !

PR #212, opened 2020-02-25
Use native threads instead of the thread helper library.

bmoyles0117:
This change results in a significant reduction of memory usage by avoiding the thread map in the thread helper. 

The thread helper is not necessary for this functionality, threads this plugin creates do not need to be synchronized with the process shutdown, as they are asynchronous processes that collect data in the background to help facilitate the enrichment of logs.

### DIFF
--- a/lib/fluent/plugin/kubernetes_metadata_watch_namespaces.rb
+++ b/lib/fluent/plugin/kubernetes_metadata_watch_namespaces.rb
@@ -35,7 +35,7 @@ module KubernetesMetadata
       # processing will be swallowed and retried. These failures /
       # exceptions could be caused by Kubernetes API being temporarily
       # down. We assume the configuration is correct at this point.
-      while thread_current_running?
+      while true
         begin
           namespace_watcher ||= get_namespaces_and_start_watcher
           process_namespace_watcher_notices(namespace_watcher)
--- a/lib/fluent/plugin/filter_kubernetes_metadata.rb
+++ b/lib/fluent/plugin/filter_kubernetes_metadata.rb
@@ -271,10 +271,10 @@ module Fluent::Plugin
         end
 
         if @watch
-          pod_thread = Thread.new(self) { |this| this.setup_pod_thread }
+          pod_thread = Thread.new(self) { |this| this.set_up_pod_thread }
           pod_thread.abort_on_exception = true
 
-          namespace_thread = Thread.new(self) { |title| this.setup_namespace_thread }
+          namespace_thread = Thread.new(self) { |title| this.set_up_namespace_thread }
           namespace_thread.abort_on_exception = true
         end
       end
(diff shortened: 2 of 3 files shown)


bmoyles0117:
Screen Shot 2020-02-25 at 1 50 46 PM

Here's a screenshot demonstrating the difference.

bmoyles0117:
Well, that was embarrassing, here's an actual screenshot of the improved memory usage.

x1YrS5Q7tT9

qingling128:
Note: This is a followup for https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/pull/204.

lib/fluent/plugin/kubernetes_metadata_watch_namespaces.rb
```diff
@@ -35,7 +35,7 @@ def set_up_namespace_thread
       # processing will be swallowed and retried. These failures /
       # exceptions could be caused by Kubernetes API being temporarily
       # down. We assume the configuration is correct at this point.
-      while thread_current_running?
+      while true
```
jcantrill: I presume the previous logic had some kind of graceful exit?  Do we have concerns with an infinite loop and graceful shutdown?  I don't know the answer but am asking the question.  Otherwise lgtm

qingling128: Yeah, the design of [thread_current_running](https://github.com/fluent/fluentd/blob/2df5ed1f7f5f2190285642b780b1fd4bc73887d8/lib/fluent/plugin_helper/thread.rb) is for graceful exit. I dug through their code a bit. It seems like an overkill for the cases we are handling here. Thus we proposed to just use the native threading.

If the watcher is healthy, we do expect it to be an infinite loop.

mszabo: We had problems with an infinite loop related to this code (not triggering the backoff). I described it in #213

jcantrill: We likely should accept this PR as is and address #213 in a subsequent change

bmoyles0117: Hold off on merging, there's actually a huge oversight on my part, I am calling a non-existent method `setup_pod_thread`, when it's supposed to be `set_up_pod_thread`, yet things are working as expected, something is either very wrong with our validation or I'm not testing what I thought I'm testing.

bmoyles0117:
Ok, we're all clear for approval now, a note for @qingling128, this reduces the memory footprint by about ~10MB, but not down to previous levels. Still worth submitting as-is.

qingling128:
@jcantrill - Could you help cut a release that includes this fix? Thanks in advance.

jcantrill:
https://rubygems.org/gems/fluent-plugin-kubernetes_metadata_filter/versions/2.4.4

qingling128:
Thanks @jcantrill !

Issue #227, opened 2020-04-13
Please publish 2.4.6 with fix for `:namespace_watch_retry_count` 


diclophis:
Hello,

Thank your for all the work so far on this plugin... we are encountering the error with `:namespace_watch_retry_count` that was recently fixed ... but we have not seen the next version published into the gem system yet (we expect to see 2.4.6 with the fix)

If there is anything I or my team can do to help with this please let me know.

Thanks

-Jon

iyurev:
+1

jcantrill:
https://rubygems.org/gems/fluent-plugin-kubernetes_metadata_filter/versions/2.4.6

PR #232, opened 2020-05-07
Lock fluent version for patch release so as not to get unexpected res…

jcantrill:
…ults

Co-authored-by: Jeff Cantrill <jcantril@redhat.chom>

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,7 +2,7 @@ PATH
   remote: .
   specs:
     fluent-plugin-kubernetes_metadata_filter (2.4.5)
-      fluentd (>= 0.14.0, < 2)
+      fluentd (>= 0.14.0, < 1.10)
       kubeclient (< 5)
       lru_redux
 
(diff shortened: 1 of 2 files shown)


jcantrill:
@qingling128 I had to rebuild my machine and having issue publishing.  Seems there are updates in the lock file so I'm asking your opinion this change seems reasonable if we are only going to do patch bump?.  Trying to resolve #227

jcantrill:
cc @richm @ewolinetz  thoughts?

richm:
> cc @richm @ewolinetz thoughts?

I'm not sure what is the general guidance for ruby gems publishers, but this seems ok to me.

qingling128:
Looks reasonable to me.

mindw:
@jcantrill, @richm - latest fluentd is 1.10.4. 1.10.0 was released march 24th. 
This change will prevent the plugin from being installed on the latest stable fluentd version. 
Just FYI - thank you for putting the time and effort on maintaining the plugin!

jcantrill:
> @jcantrill, @richm - latest fluentd is 1.10.4. 1.10.0 was released march 24th.
> This change will prevent the plugin from being installed on the latest stable fluentd version.
> Just FYI - thank you for putting the time and effort on maintaining the plugin!

@mindw Understand completely but this is intentional given a patch release of this plugin IMO should not bump to require a minor version change of fluent.  Please consider submitting a patch that would bump the fluent dependency so it can be used on 1.10

Issue #226, opened 2020-04-08
Error while watching pods: too old resource version with v2.4.5


raja-gola:
Error while watching pods: too old resource version: 23630883 (23632006) (RuntimeError)

complete stacktrace
```
#<Thread:0x00007f53f192fb70@/usr/local/bundle/gems/fluent-plugin-kubernetes_metadata_filter-2.4.5/lib/fluent/plugin/filter_kubernetes_metadata.rb:274 run> terminated with exception (report_on_exception is true):
/usr/local/bundle/gems/fluent-plugin-kubernetes_metadata_filter-2.4.5/lib/fluent/plugin/kubernetes_metadata_watch_pods.rb:43:in `rescue in set_up_pod_thread': undefined method `<' for nil:NilClass (NoMethodError)
        from /usr/local/bundle/gems/fluent-plugin-kubernetes_metadata_filter-2.4.5/lib/fluent/plugin/kubernetes_metadata_watch_pods.rb:38:in `set_up_pod_thread'
        from /usr/local/bundle/gems/fluent-plugin-kubernetes_metadata_filter-2.4.5/lib/fluent/plugin/filter_kubernetes_metadata.rb:274:in `block in configure'
/usr/local/bundle/gems/fluent-plugin-kubernetes_metadata_filter-2.4.5/lib/fluent/plugin/kubernetes_metadata_watch_pods.rb:133:in `block in process_pod_watcher_notices': Error while watching pods: too old resource version: 23630883 (23632006) (RuntimeError)
        from /usr/local/bundle/gems/kubeclient-4.6.0/lib/kubeclient/watch_stream.rb:28:in `block in each'
        from /usr/local/bundle/gems/http-4.4.1/lib/http/response/body.rb:37:in `each'
        from /usr/local/bundle/gems/kubeclient-4.6.0/lib/kubeclient/watch_stream.rb:25:in `each'
        from /usr/local/bundle/gems/fluent-plugin-kubernetes_metadata_filter-2.4.5/lib/fluent/plugin/kubernetes_metadata_watch_pods.rb:110:in `process_pod_watcher_notices'
        from /usr/local/bundle/gems/fluent-plugin-kubernetes_metadata_filter-2.4.5/lib/fluent/plugin/kubernetes_metadata_watch_pods.rb:40:in `set_up_pod_thread'
[...]

PR #243, opened 2020-07-10
Add watch 410 Gone response handling

Ghazgkull:
This PR addresses Issue #226 by adding handling for 410 Gone responses to the pod watch handling in this plugin.

I've included a unit test which verifies the expected behavior.

### DIFF
--- a/lib/fluent/plugin/kubernetes_metadata_watch_pods.rb
+++ b/lib/fluent/plugin/kubernetes_metadata_watch_pods.rb
@@ -19,6 +19,13 @@
 require_relative 'kubernetes_metadata_common'
 
 module KubernetesMetadata
+
+  class GoneError < StandardError
+    def initialize(msg="410 Gone")
+      super
+    end
+  end
+
   module WatchPods
 
     include ::KubernetesMetadata::Common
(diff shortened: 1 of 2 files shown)


Ghazgkull:
Please note that after some reflection and further reading, I pushed a new commit which handles the error in a much better way. The approach here is to throw a typed exception (GoneError) which can be caught in `set_up_pod_thread` in order to restart `get_pods_and_start_watcher`. This means we'll go back out and get the latest resourceVersions, which should allow us to gracefully recover from the 410.

jcantrill:
@Ghazgkull  this LGTM.  Do you have anything you can post ouside of the unit test to confirm this addresses the issue?  Have you  built and run it manually?

Ghazgkull:
@jcantrill I'd love to do more validation, but I would need some guidance.

Prior to submitting this PR, my experience with fluentd is limited to deploying the fluentd-elasticsearch helm chart to my clusters. Trying to imagine how I'd be able to test this out in a deployment feels like it might be a major undertaking. I imagine I'd have to:
1. Build and publish this plugin somewhere. I haven't worked in Ruby in years so I don't remember anything about where gems are published and how to publish pre-release versions. I'd have to figure that out.
2. Fork the elasticsearch-fluentd repo and bump the version of this plugin in their Gemfile. Then build and publish that Docker image somewhere. That doesn't sound too bad as long as their project can be built without a lot of setup.
3. Update the elasticsearch-fluentd helm chart to use that image. That bit is trivial because the helm chart allows to specify what image you want deployed without modifying the chart.

jcantrill:
There is probably an easier way to test it but further reviewing the changes, I don't believe we are any worse off then before and the unit tests confirm behavior.

Issue #246, opened 2020-07-16
Add 410 Gone handling for namespace watch


Ghazgkull:
Similar to the issue we fixed last week for pod watches (#226), we should also gracefully handle 410 Gone errors on the namespace watch.

Ghazgkull:
@jcantrill I've submitted a PR which fixes this issue, https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/pull/247.

If this PR looks good to you and we can get a new version of this plugin pushed soon, I already have a PR open for the fluend-es-image in the Kubernetes project and I'd be happy to get this fix included there so it's available to everyone. (https://github.com/kubernetes/kubernetes/pull/93116/)

jcantrill:
released in https://rubygems.org/gems/fluent-plugin-kubernetes_metadata_filter/versions/2.5.2

PR #247, opened 2020-07-16
Add 410 Gone handling for namespace watch

Ghazgkull:
@jcantrill This is a follow-up to the PR I submitted last week to add handling for 410 Gone responses on the pod watch.

This PR makes two changes:
1. It adds the same handling for the namespace watch.
2. I changed the log level of the message from debug to info. Based on my experience running this in the wild, this is a message that we should print with the default log level. It's important for folks to know this is happening and that we're handling it.

This fixes issue #246

### DIFF
--- a/lib/fluent/plugin/kubernetes_metadata_common.rb
+++ b/lib/fluent/plugin/kubernetes_metadata_common.rb
@@ -19,6 +19,12 @@
 module KubernetesMetadata
   module Common
 
+    class GoneError < StandardError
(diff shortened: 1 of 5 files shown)


Ghazgkull:
Please hold on this. I'm adding more tests and will update the PR with a new commit shortly.

Ghazgkull:
New commit is attached, which includes tests for the controlling thread loop as well.

PR #250, opened 2020-08-31
run tests with bundler + allow running a single test file

grosser:
@fabric8io

also added `TEST_NAME='with records from journald and docker & kubernetes metadata, alternate form' ruby test/plugin/test_filter_kubernetes_metadata.rb`so a single test can be run

### DIFF
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -16,6 +16,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+require 'bundler/setup'
 require 'codeclimate-test-reporter'
 SimpleCov.start do
   formatter SimpleCov::Formatter::MultiFormatter.new [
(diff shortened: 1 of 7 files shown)


test/helper.rb
```diff
@@ -31,8 +32,14 @@
...
 require 'minitest/autorun'
-require 'webmock/test_unit'
 require 'vcr'
+require 'ostruct'
+require 'fluent/plugin/filter_kubernetes_metadata'
+require 'fluent/test/driver/filter'
+require 'kubeclient'
+
+require 'webmock/test_unit'
+WebMock.disable_net_connect!
```
grosser, line 42: this modifies the global environment, so needs to be in the helper or running tests in a different order / single test leads to weird behavior

jcantrill (approved):

PR #254, opened 2020-09-01
use more efficient hashes when talking to kubernetes

grosser:
more details here https://github.com/abonas/kubeclient/issues/251

### DIFF
--- a/lib/fluent/plugin/kubernetes_metadata_watch_namespaces.rb
+++ b/lib/fluent/plugin/kubernetes_metadata_watch_namespaces.rb
@@ -92,12 +92,12 @@ module KubernetesMetadata
         resource_version: '0'  # Fetch from API server.
       }
       namespaces = @client.get_namespaces(options)
-      namespaces.each do |namespace|
-        cache_key = namespace.metadata['uid']
+      namespaces[:items].each do |namespace|
+        cache_key = namespace[:metadata][:uid]
         @namespace_cache[cache_key] = parse_namespace_metadata(namespace)
         @stats.bump(:namespace_cache_host_updates)
       end
-      options[:resource_version] = namespaces.resourceVersion
+      options[:resource_version] = namespaces[:metadata][:resourceVersion]
       @client.watch_namespaces(options)
     end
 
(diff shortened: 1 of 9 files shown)


jcantrill:
@grosser started through your PRs from oldest... looks like we need a rebase here

grosser:
rebased!

lib/fluent/plugin/kubernetes_metadata_watch_namespaces.rb
```diff
@@ -92,14 +92,13 @@ def get_namespaces_and_start_watcher
...
-        cache_key = namespace.metadata['uid']
+      namespaces[:items].each do |namespace|
+        cache_key = namespace[:metadata][:uid]
         @namespace_cache[cache_key] = parse_namespace_metadata(namespace)
         @stats.bump(:namespace_cache_host_updates)
       end
-      options[:resource_version] = namespaces.resourceVersion
-      watcher = @client.watch_namespaces(options)
-      watcher
+      options[:resource_version] = namespaces[:metadata][:resourceVersion]
```
cben, line 100: :+1: this a special case where the change is not 1-to-1:

> With default (as: :ros) return format, the returned object acts like an array of the individual pods, but also supports a .resourceVersion method.

> With :parsed and :parsed_symbolized formats, the returned data structure matches kubernetes list structure: it's a hash containing metadata and items keys, the latter containing the individual pods.

test/plugin/test_watch_pods.rb
```diff
@@ -24,140 +24,141 @@ class DefaultPodWatchStrategyTest < WatchTest
...
-                                                           'name' => 'bar',
-                                                           'image' => 'foo',
-                                                         }]
-                                      }
-                                    }),
-         ])
-       @created = OpenStruct.new(
+       @initial = {
+         kind: 'PodList',
+         metadata: {resourceVersion: '123'}, # TODO: resourceVersion is always blank
```
cben, line 29: What do you mean? In test?  In real life?

test/plugin/watch_test.rb
```diff
@@ -36,24 +36,21 @@ def thread_current_running?
...
+      {items: [], metadata: {resourceVersion: '12345'}}
     end
     def @client.get_pods(options = {})
-      self
+      {items: [], metadata: {resourceVersion: '12345'}}
     end
 
     @exception_raised = OpenStruct.new
-    def @exception_raised.each
+    def @exception_raised.[](_)
```
cben, line 53: This one is a bit strange.  It's used in 2 tests like this:
```
        @client.stub :watch_namespaces, [[@created, @exception_raised]] do
```
Is the array used to simulate the watch object .each first yielding a value, then raising an exception? But instead of the watch object raising, it yields an "explosive" value that raises when it gets accessed?

jcantrill:
I'll defer to responding to @cben before merging.

test/plugin/test_watch_pods.rb
grosser: fixed, was a leftover comment from debugging

test/plugin/watch_test.rb
grosser: made this a bit more obvious ... it was meant to cause an exception when being used, since "correct" notices do not blow up, but any undefined behavior is rescued

grosser:
@cben looks good ?

cben (approved):
LGTM :100:

jcantrill:
@grosser looks like there is a test failure. please investigate

grosser:
rebuild fixed it ... some random error :)

grosser:
@jcantrill 👋

PR #255, opened 2020-09-01
bump required ruby version to what the readme says

grosser:


fluent-plugin-kubernetes_metadata_filter.gemspec
```diff
@@ -13,11 +13,8 @@ Gem::Specification.new do |gem|
   gem.license       = "Apache-2.0"
 
   gem.files         = `git ls-files`.split($/)
-  gem.executables   = gem.files.grep(%r{^bin/}) { |f| File.basename(f) }
-  gem.test_files    = gem.files.grep(%r{^(test|spec|features)/})
-  gem.require_paths = ["lib"]
 
-  gem.required_ruby_version = '>= 2.1.0'
+  gem.required_ruby_version = '>= 2.5.0'
```
jcantrill, line 17: This causes a test failure which leads me to believe either we should drop this job https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/blob/master/.circleci/config.yml#L25 or maybe set it to 2.5.0

grosser:
replaced 2.4 with 2.7 :)

Issue #251, opened 2020-08-31
formatting does not match ruby standards


grosser:
would you accept a PR that makes everything be 2-space indent etc minor fixes ?

jcantrill:
Yes.  Might we add a linter or pre-commit hook to format such that it meets ruby standards?

grosser:
https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/pull/264

PR #264, opened 2020-10-08
add rubocop

grosser:
fixes https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/issues/251#issuecomment-705268717


grosser:
https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/pull/265 with just the tests to make sure it does not break both and ends up green by accident

.circleci/config.yml
```diff
@@ -17,7 +17,7 @@ missingdeps: &missingdeps
 
 test: &test
   name: Test bundle
-  command: bundle exec rake test
+  command: bundle exec rake --trace
```
jcantrill, line 20: Why this change?

jcantrill (commented):
Aside from one question it looks like formatting changes which is to be expected

.circleci/config.yml
grosser: it makes it run rubocop and the test
added --trace so it's obvious if test failed or rubocop, it will show the task name

jcantrill:
@grosser please rebase if appropriate or close

jcantrill:
Close in favor of #278

PR #276, opened 2021-01-23
fix #263: use filter in lieu of filter_stream as recommended optimization

### LINKED ISSUE #263: change filter_stream to filter
I wrote some other plugins and alway used filter, so wanted to check which is best and https://docs.fluentd.org/filter#filter-chain-optimization says to not use filter_stream ... is there any reason to keep it or can it be changed as long as the tests pass ?

jcantrill:
fix #263: use filter in lieu of filter_stream as recommended optimiza

### DIFF
--- a/test/plugin/test_filter_kubernetes_metadata.rb
+++ b/test/plugin/test_filter_kubernetes_metadata.rb
@@ -156,15 +156,6 @@ class KubernetesMetadataFilterTest < Test::Unit::TestCase
       d.filtered.map(&:last)
     end
 
-    test 'nil event stream' do
-      # not certain how this is possible but adding test to properly
-      # guard against this condition we have seen - test for nil,
-      # empty, no empty method, not an event stream
-      plugin = create_driver.instance
-      plugin.filter_stream('tag', nil)
-      plugin.filter_stream('tag', Fluent::MultiEventStream.new)
-    end
-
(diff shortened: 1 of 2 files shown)


jcantrill:
cc @ewolinetz @alanconway @vimalk78  are takers for review?

alanconway (commented):
The change looks OK, but I dont' see the changes to callers of filter_stream

jcantrill:
> The change looks OK, but I dont' see the changes to callers of filter_stream

@alanconway I believe the callers are unknown to most users as it handled internally by fluentd when you define use of the plugin via configuration.  It is fluentd that makes the optimization.

test/plugin/test_filter_kubernetes_metadata.rb
```diff
@@ -145,18 +145,9 @@ def emit_with_tag(tag, msg = {}, config = '
...
-      # not certain how this is possible but adding test to properly
-      # guard against this condition we have seen - test for nil,
-      # empty, no empty method, not an event stream
-      plugin = create_driver.instance
-      plugin.filter_stream('tag', nil)
-      plugin.filter_stream('tag', Fluent::MultiEventStream.new)
+      d.filtered_records
     end
 
     test 'inability to connect to the api server handles exception and doensnt block pipeline' do
```
syedriko, line 153: s/doensnt/doesn't

PR #280, opened 2021-03-06
Add podIP to metadata filter

vkperumal:
This PR  adds podIP to metadata. This is helpful with correlating with access logs.


vkperumal:
@jcantrill all the test cases are fixed which ran successfully locally

PR #283, opened 2021-04-23
Add capability to configure a simple cache strategy

jcantrill:
This PR introduces configuration capability to configure a simple caching strategy:
* Makes no effort to try to provide labels from deleted pods which no longer exist

cc @eranra this would be worth to build into an image and run through the benchmarks

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -39,3 +39,42 @@ task :headers do
   command_line = CopyrightHeader::CommandLine.new(args)
   command_line.execute
 end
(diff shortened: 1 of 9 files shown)


jcantrill:
Closing given with the merge of https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/pull/317
* the pod uuid is available as part of the file metadata
* Reduced need for an easier mechanism

PR #284, opened 2021-04-26
fix #257. Add warning log when 'K8S_NODE_NAME' is not set

### LINKED ISSUE #257: make K8S_NODE_NAME required when in kubernetes
not setting K8S_NODE_NAME results in slow/big requests to `pods` list, which can harm the api server and the stability of the whole cluster, would you accept a PR to make it required when in kubernetes (via env var detection for example) ?
atm someone just installing the plugin without reading the whole readme can stumble into this bug and it's pretty nasty :(
... can have an opt-out for weird setups

jcantrill:
This PR:

* fixes #257 by issuing a warning message when the node name is not set
* warning chosen over requirement to address backwards compatibility

cc @alanconway

### DIFF
--- a/lib/fluent/plugin/filter_kubernetes_metadata.rb
+++ b/lib/fluent/plugin/filter_kubernetes_metadata.rb
@@ -253,6 +253,10 @@ module Fluent::Plugin
         end
 
         if @watch
+          if ENV['K8S_NODE_NAME'].nil? || ENV['K8S_NODE_NAME'].strip.empty?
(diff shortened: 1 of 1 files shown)


alanconway (approved):
I think we might be able to do a more efficient job with a type watch (all Pods) rather than a long list of instance watches, but it would require some investigation.

/lgtm

jcantrill:
@alanconway   the functionality already exists within this plugin to watch pods and restrict that watch to a given node.  That restriction is passed in by setting K8S_NODE_NAME.  This means the number of pods will be very small.  The original ask was to make the variable required but I don't think that will guarantee backwards compatibility.  This change will warn the user that its not efficient.

PR #295, opened 2021-06-10
chore(gems): peg kubeclient to >= 4.0.0

Cryptophobia:
- peg kubeclient gem to 4.x.x versions
 
 Signed-off-by: Anton Ouzounov <aouzounov@vmware.com>
 
 Recently had a problem where `bundle install` chose to install gem version `1.0.0` for `kubeclient` gem. I am not exactly sure what exactly caused this but I know that `fluent-plugin-kubernetes_metadata_filter` is the only gem that requires `kubeclient` when reversing the gem dependencies. But I believe that this small change may solve it.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -3,7 +3,7 @@ PATH
   specs:
     fluent-plugin-kubernetes_metadata_filter (2.7.1)
       fluentd (>= 0.14.0, < 1.14)
-      kubeclient (< 5)
+      kubeclient (>= 4.0.0, < 5.0.0)
       lru_redux
 
 GEM
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,7 +2,7 @@ PATH
   remote: .
   specs:
     fluent-plugin-kubernetes_metadata_filter (2.7.0)
-      fluentd (>= 0.14.0, < 1.13)
+      fluentd (>= 0.14.0, < 1.14)
       kubeclient (< 5)
       lru_redux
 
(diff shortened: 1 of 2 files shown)


Cryptophobia:
@jcantrill , had a weird dependency error where bundler installed `kubeclient (1.0.0)` instead of the latest `kubeclient (4.9.2)`. I think if we define stricter versioning in the .gemspec file when pulling the gem from Rubygems, it should force bundler to use a later version of `kubeclient`.

cben:
Makes sense :+1:, kubeclient had some breaking changes between major releases (we try to follow semver there), so allowing more than one major version is risky.

Gemfile.lock
```diff
@@ -3,7 +3,7 @@ PATH
   specs:
     fluent-plugin-kubernetes_metadata_filter (2.7.1)
       fluentd (>= 0.14.0, < 1.14)
-      kubeclient (< 5)
+      kubeclient (>= 4.0.0, < 5.0.0)
```
jcantrill, line 6: Do we have any testing outside units the version bump is not causing any regression? Can we document what version of kube we know to work?

Cryptophobia: This is the error when initializing the kubeclient when we pulled v1.0.0 instead of >= v4.0.0 using bundler:

```
2021-06-10 00:50:53 +0000 [debug]: [filter_kube_metadata] Creating K8S client
/usr/local/lib/ruby/gems/2.7.0/gems/kubeclient-1.0.0/lib/kubeclient.rb:34:in `initialize': unknown keyword: :as (ArgumentError)
```

The `:as` option was introduced in the kubeclient initialization gem in [kubeclient/lib/kubeclient/common.rb](https://github.com/abonas/kubeclient/blob/e37a8029275a6032a6f9b641b3800601ff1a2a0b/lib/kubeclient/common.rb) during v3.0.0 tag. I would expect everything below <= 3.0.0 of kubeclient gem to break because of this change to initialization. I am no ruby expert though.

Cryptophobia: I scouted the tags in versions and the initialization function has not change much from v4.0.0 to v4.9.2 so it makes me think that any version over 4.0.0 would work.

Cryptophobia:
@jcantrill , any ideas if anything else needs to be done for this PR to merge? Any particular testing that we would want to perform?

jcantrill:
Closing in favor of https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/pull/304

Cryptophobia:
Thank you @jcantrill ! I can delete my fork now. :)

PR #304, opened 2021-09-01
fix #295. Fix min/max version of kubeclient dep

jcantrill:
Supersedes https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/pull/295 to limit kube client version


cben:
LGTM

PR #307, opened 2021-09-17
[Fix #306] Add de_slash with de_slash_separator feature for Datadog Logs support

### LINKED ISSUE #306: Datadog logs facets don't support `/` in paths
Datadog logs facets don't support slashes and/or dots in paths.

Kubernetes recommended labels come with slashes and dots: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

Screenshot 2021-09-17 at 16 09 45

We've noticed the current `de_dot` feature enabled by default, for Elasticsearch.

Would be useful to have a `de_slash` feature too, so that we can easily configure the plugin to be Datadog ready.

rymir:
This pull request mirrors the `de_dot` feature, adding an additional `de_slash` functionality, so that Datadog configuration can be supported when using recommended kubernetes labels https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/


lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -58,6 +58,8 @@ class KubernetesMetadataFilter < Fluent::Plugin::Filter
     config_param :secret_dir, :string, default: '/var/run/secrets/kubernetes.io/serviceaccount'
     config_param :de_dot, :bool, default: true
     config_param :de_dot_separator, :string, default: '_'
+    config_param :de_slash, :bool, default: true
```
jcantrill, line 61: This should default to `false` to ensure backwards compatibility

jcantrill (changes requested):

lib/fluent/plugin/filter_kubernetes_metadata.rb
rymir: ✅

jcantrill (approved):

PR #311, opened 2021-10-01
fix #310. Handle case where there is no container_status

### LINKED ISSUE #310: undefined method `each' for nil:NilClass for new pods
with the 2.9.0 version one gets lots of warning outpuds pretty much every time a new pod starts:

    fluentd-cluster-logging-pzw4m fluentd-logging 2021-09-28 07:46:42 +0000 [warn]: parsing container meta information failed for: testing-qa/-test-6d8cbd8555-vx55l: undefined method `each' for nil:NilClass

and

    fluentd-cluster-logging-hldcd fluentd-logging 2021-09-28 08:09:34 +0000 [warn]: #0 parsing container meta information failed for: ams-testing-qa/-qa-5486cbfcfd-d468t: undefined method `sub' for nil:NilClass

this is probably no regression as the message was changed to warning in 2.9.0
the error occurs here:

https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/commit/d96622ed23c8ec6ad9c52a4c4d9c2bbf4dfdbd3d#diff-06faf62948d283e7de8978dee03a0f28ef58256693344f90ea330728f29f1fb3R74

jcantrill:
This issue fixes #310 by handling the case where returned pod meta does not include container_status or other bit. (e.g. pod initializing)

@juliantaylor  Care to review?

cc @alanconway @vimalk78

### DIFF
--- a/lib/fluent/plugin/kubernetes_metadata_common.rb
+++ b/lib/fluent/plugin/kubernetes_metadata_common.rb
@@ -79,7 +79,7 @@ module KubernetesMetadata
       begin
         pod_object[:status][:containerStatuses].each do |container_status|
           # get plain container id (eg. docker://hash -> hash)
-          container_id = container_status[:containerID].sub(%r{^[-_a-zA-Z0-9]+://}, '')
+          container_id = container_status[:containerID] ? container_status[:containerID].sub(%r{^[-_a-zA-Z0-9]+://}, '') : container_status[:name]
           container_meta[container_id] = if @skip_container_metadata
                                            {
                                              'name' => container_status[:name]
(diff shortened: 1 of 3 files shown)


lib/fluent/plugin/kubernetes_metadata_common.rb
```diff
@@ -79,7 +79,7 @@ def parse_pod_metadata(pod_object)
       begin
         pod_object[:status][:containerStatuses].each do |container_status|
```
juliantaylor, line 80: containerStatuses can be nil when the pod is e.g. in Pending state

juliantaylor: ah that is caught by the if on the end? not very familiar with ruby

juliantaylor:
tested patch with in the same tests from the issue (new pods, pending pods), no warnings and the metadata is filled correctly for running pods.

Issue #323, opened 2022-01-31
k8s 1.22 - bearer tokens from mounts needs to be refreshed


kjetijor:
k8s 1.22 starts enforcing lifetime for the bearer tokens mounted in to the pod, this means they'll go stale after an hour, and we'll start getting rejected by the k8s api servers.

See: https://github.com/abonas/kubeclient/issues/530

cben:
As I wrote on #328, this was not addressed.

HarleyB123:
Any updates on when this will be fixed?

jcantrill:
@cben Based upon your original response in #328, I agree we should continue to pursue your proposed changes.  Please review my code comments.

prashil-g:
Any update on this?

PettitWesley:
Here is another way of doing it, reactively refresh the token when you get an Unauthorized error: https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/pull/337

jcantrill:
Released https://rubygems.org/gems/fluent-plugin-kubernetes_metadata_filter/versions/2.11.1

PR #327, opened 2022-03-16
WIP: Use bearer token file and periodically recreate client

emiran-orange:
Recreate k8s client periodically.
When kubeclient 5 is released, faraday_client will just have to be unset to be recreated.

Fixes https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/issues/323

### DIFF
--- a/lib/fluent/plugin/filter_kubernetes_metadata.rb
+++ b/lib/fluent/plugin/filter_kubernetes_metadata.rb
@@ -161,6 +161,20 @@ module Fluent::Plugin
       @prev_time = Time.now
     end
 
+    def periodically_read_bearer_token_file
+      token_read_period = 10
+      while true
+        log.debug "Sleeping for #{token_read_period}"
+        sleep(token_read_period)
+        log.debug "Updating bearer_token"
+        # This will be enough with kubeclient 5
+        # @client.instance_variable_set(:@faraday_client, nil)
+        # until then
+        @client.bearer_token(File.read(@bearer_token_file))
+        # @client.instance_variable_set(:@rest_client, nil)
+      end
+    end
+
     def configure(conf)
       super
 
@@ -191,7 +205,7 @@ module Fluent::Plugin
       @namespace_cache = LruRedux::TTL::ThreadSafeCache.new(@cache_size, @cache_ttl)
 
       @tag_to_kubernetes_name_regexp_compiled = Regexp.compile(@tag_to_kubernetes_name_regexp)
-      
+
       @container_name_to_kubernetes_regexp_compiled = Regexp.compile(@container_name_to_kubernetes_regexp)
 
       # Use Kubernetes default service account if we're in a pod.
@@ -255,8 +269,7 @@ module Fluent::Plugin
         auth_options = {}
 
         if present?(@bearer_token_file)
-          bearer_token = File.read(@bearer_token_file)
-          auth_options[:bearer_token] = bearer_token
+          auth_options[:bearer_token_file] = @bearer_token_file
         end
 
         log.debug 'Creating K8S client'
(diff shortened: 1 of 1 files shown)


lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -161,6 +161,19 @@ def initialize
       @prev_time = Time.now
     end
 
+    def periodically_recreate_k8s_client
+      token_read_period = 10
+      while true
+        log.debug "Sleeping for #{token_read_period}"
+        sleep(token_read_period)
```
jcantrill, line 168: This seems like it will pause all processing.  I believe a better implementation would be to follow the pattern of https://github.com/ManageIQ/kubeclient/pull/532/files

emiran-orange: @jcantrill 
I'm not really skilled in threads and in ruby, how that having a sleep in a thread (not thread.sleep) pauses all processing ?

Having read about "||=" statement which I didn't know, `@faraday_client ||= create_faraday_client` seems to be equivalent to `@faraday_client || @faraday_client = create_faraday_client` not `@faraday_client = @faraday_client || create_faraday_client` which means `@faraday_client` needs to be set to nil/false to actually be recreated. Have I been misguided ?

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -161,6 +161,20 @@ def initialize
...
+        log.debug "Updating bearer_token"
+        # This will be enough with kubeclient 5
+        # @client.instance_variable_set(:@faraday_client, nil)
+        # until then
+        @client.bearer_token(File.read(@bearer_token_file))
+        # @client.instance_variable_set(:@rest_client, nil)
```
cben, line 174: All these options rely on knowing the internals of kubeclient.  
How hard it is to recreate the whole `Kubeclient::Client.new(...)`?  I see that's what #326 was going for :+1:

Also both `@rest_client` (kubeclient 4) and `@faraday_client` (5) only cover CRUD actions but not watches, doesn't this plugin rely on watches?

- I don't think an open watch socket will benefit from any of these approaches, but OTOH it's already open and doesn't need auth, right?  At some point server might disconnect it and you'll open new watch with fresh auth.

I guess that's why you closed this and helped improve the file re-reading in kubeclient itself. (Thanks for that!)  

But I'm afraid I'm not yet close to releasing kubeclient 5.  I'm working through a backlog in my spare time but it might take a month at least (and that's not a commitment!)
If you need this soon, I recommend the #326 approach.

emiran-orange: @cben Thanks for sharing this.
I thought recreating the client in #326 was a bit brutal (and I had trouble with the failing tests ...) but I agree with the fact that relying on kubeclient internals may not be the best way (and it still doesn't work actually).
I think I kinda fixed the tests in #326.
This PR and #326 share the same periodic approach, I'll wait until @jcantrill answers https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/pull/327#discussion_r828193326 to close this one and re-open #326

lib/fluent/plugin/filter_kubernetes_metadata.rb
jcantrill: I don't believe there is any reason to add threading to monitor the client.  It's expiration should happen infrequently enough (e.g. 60 min) that introducing logic to recreate on failure should be sufficient.  I believe the pattern established in https://github.com/ManageIQ/kubeclient/pull/532 is sufficient

emiran-orange: Actually the pattern is not sufficient when `watch` is false as no metadata will be retrieved.
BTW the periodic approach is the one used in kubernetes client-go to refresh token.

PR #328, opened 2022-03-23
Use bearer token file and periodically recreate client

### LINKED ISSUE #323: k8s 1.22 - bearer tokens from mounts needs to be refreshed

emiran-orange:
Recreate k8s client periodically.
As discussed in #327 this seems a better approach than waiting for the token to expire.

NB: I couldn't reopen #326 hence this new PR (sorry for the mess)

Fixes #323

### DIFF
--- a/lib/fluent/plugin/filter_kubernetes_metadata.rb
+++ b/lib/fluent/plugin/filter_kubernetes_metadata.rb
@@ -161,6 +161,61 @@ module Fluent::Plugin
       @prev_time = Time.now
     end
 
+    def create_k8s_client
+        ssl_options = {
+          client_cert: present?(@client_cert) ? OpenSSL::X509::Certificate.new(File.read(@client_cert)) : nil,
+          client_key: present?(@client_key) ? OpenSSL::PKey::RSA.new(File.read(@client_key)) : nil,
+          ca_file: @ca_file,
+          verify_ssl: @verify_ssl ? OpenSSL::SSL::VERIFY_PEER : OpenSSL::SSL::VERIFY_NONE
+        }
+
+        if @ssl_partial_chain
+          # taken from the ssl.rb OpenSSL::SSL::SSLContext code for DEFAULT_CERT_STORE
+          require 'openssl'
+          ssl_store = OpenSSL::X509::Store.new
+          ssl_store.set_default_paths
+          flagval = if defined? OpenSSL::X509::V_FLAG_PARTIAL_CHAIN
+                      OpenSSL::X509::V_FLAG_PARTIAL_CHAIN
+                    else
+                      # this version of ruby does not define OpenSSL::X509::V_FLAG_PARTIAL_CHAIN
+                      0x80000
+                    end
+          ssl_store.flags = OpenSSL::X509::V_FLAG_CRL_CHECK_ALL | flagval
+          ssl_options[:cert_store] = ssl_store
+        end
+
+        auth_options = {}
+
+        if present?(@bearer_token_file)
+          auth_options[:bearer_token_file] = @bearer_token_file
+        end
+
+        log.debug 'Creating K8S client'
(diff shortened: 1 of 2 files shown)


jcantrill:
I don't think any of this change is required other then to bump the library and modify it to use bearer token path instead of the token directly.  Not This lib depends upon https://github.com/ManageIQ/kubeclient which is where https://github.com/ManageIQ/kubeclient/pull/532 was merged in 2022-Feb-13

cben:
No, kubeclient only implemented token refresh on master branch, on top of `faraday` + `http`, which will one day be released as 5.0 (can't really commit to timing but at current rate of me rarely working on it I'd say at least couple months :disappointed:).

None of the 4.y.z releases have token refresh.  Options if you want it soon:
- resurrect this.
- work on backporting the refresh to kubeclient `v4.y` branch.  It uses a different `rest-client` gem (+ same `http` for watches), so it's not a trivial textual backport.

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -161,6 +161,61 @@ def initialize
...
+        end
+
+        auth_options = {}
+
+        if present?(@bearer_token_file)
+          auth_options[:bearer_token_file] = @bearer_token_file
+        end
+
+        log.debug 'Creating K8S client'
+        @client = Kubeclient::Client.new(
```
jcantrill, line 194: Since it is possible this method executes periodically on a timer, seems we may require some mutex to ensure we are not recreating the client in the middle of trying to use it to access the API server

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -161,6 +161,61 @@ def initialize
...
+          require_relative @test_api_adapter.underscore
+          @client.extend(eval(@test_api_adapter))
+        end
+    end
+
+    def periodically_recreate_k8s_client
+      token_read_period = 60
+      while true
+        log.debug "Sleeping for #{token_read_period}"
+        sleep(token_read_period)
```
jcantrill, line 213: I don't know the answer but asking the question:  is there a need to "pass" instead of or in conjunction with "sleep" or ar they the same

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -161,6 +161,61 @@ def initialize
       @prev_time = Time.now
     end
 
+    def create_k8s_client
```
jcantrill, line 164: Consider moving all this code into a module to reduce the complexity of the main filter file

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -230,56 +285,18 @@ def configure(conf)
...
+        create_k8s_client
         begin
           @client.api_valid?
         rescue KubeException => e
           raise Fluent::ConfigError, "Invalid Kubernetes API #{@apiVersion} endpoint #{@kubernetes_url}: #{e.message}"
         end
 
+        if present?(@bearer_token_file)
+          k8s_client_thread = Thread.new(self, &:periodically_recreate_k8s_client)
+          k8s_client_thread.abort_on_exception = true
```
jcantrill, line 297: Will this abort the entire fluent process?  I have concerns of some failure where fluent will continue to process messages but will render this plugin completely unusable with no immediate indicator for admins

PettitWesley:
Here is another way of doing it, reactively refresh the token when you get an Unauthorized error: https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/pull/337

PR #337, opened 2022-06-09
Refresh k8s client on 'Unathorized' exceptions

PettitWesley:
Signed-off-by: Wesley Pettit <wppttt@amazon.com>

#323

### DIFF
--- a/lib/fluent/plugin/filter_kubernetes_metadata.rb
+++ b/lib/fluent/plugin/filter_kubernetes_metadata.rb
@@ -121,6 +121,12 @@ module Fluent::Plugin
     rescue StandardError => e
       @stats.bump(:pod_cache_api_nil_error)
       log.debug "Exception '#{e}' encountered fetching pod metadata from Kubernetes API #{@apiVersion} endpoint #{@kubernetes_url}"
+      if e.message == "Unauthorized"
+        @client = nil
+        # recreate client to refresh token
+        log.info("Re-creating Kubernetes API Client to refresh auth bearer token.")
+        create_client()
+      end
       {}
     end
 
@@ -153,12 +159,20 @@ module Fluent::Plugin
     rescue StandardError => e
       @stats.bump(:namespace_cache_api_nil_error)
       log.debug "Exception '#{e}' encountered fetching namespace metadata from Kubernetes API #{@apiVersion} endpoint #{@kubernetes_url}"
+      if e.message == "Unauthorized"
+        @client = nil
+        # recreate client to refresh token
+        log.info("Re-creating Kubernetes API Client to refresh auth bearer token.")
+        create_client()
+      end
       {}
     end
 
     def initialize
       super
       @prev_time = Time.now
+      @ssl_options = {}
+      @auth_options = {}
     end
 
     def configure(conf)
@@ -230,7 +244,7 @@ module Fluent::Plugin
       end
 
       if present?(@kubernetes_url)
-        ssl_options = {
+        @ssl_options = {
           client_cert: present?(@client_cert) ? OpenSSL::X509::Certificate.new(File.read(@client_cert)) : nil,
           client_key: present?(@client_key) ? OpenSSL::PKey::RSA.new(File.read(@client_key)) : nil,
           ca_file: @ca_file,
@@ -249,24 +263,14 @@ module Fluent::Plugin
                       0x80000
                     end
           ssl_store.flags = OpenSSL::X509::V_FLAG_CRL_CHECK_ALL | flagval
-          ssl_options[:cert_store] = ssl_store
+          @ssl_options[:cert_store] = ssl_store
         end
-
-        auth_options = {}
 
         if present?(@bearer_token_file)
-          bearer_token = File.read(@bearer_token_file)
-          auth_options[:bearer_token] = bearer_token
+          @auth_options[:bearer_token_file] = @bearer_token_file
         end
 
-        log.debug 'Creating K8S client'
-        @client = Kubeclient::Client.new(
-          @kubernetes_url,
-          @apiVersion,
-          ssl_options: ssl_options,
-          auth_options: auth_options,
-          as: :parsed_symbolized
-        )
+        create_client()
 
         if @test_api_adapter
           log.info "Extending client with test api adaper #{@test_api_adapter}"
@@ -303,6 +307,17 @@ module Fluent::Plugin
       rescue RegexpError => e
         log.error "Error: invalid regular expression in annotation_match: #{e}"
       end
+    end
+
+    def create_client()
+      log.debug 'Creating K8S client'
+      @client = Kubeclient::Client.new(
+        @kubernetes_url,
+        @apiVersion,
+        ssl_options: @ssl_options,
+        auth_options: @auth_options,
+        as: :parsed_symbolized
+      )
     end
 
     def get_metadata_for_record(namespace_name, pod_name, container_name, cache_key, create_time, batch_miss_cache, docker_id)
--- a/lib/fluent/plugin/filter_kubernetes_metadata.rb
+++ b/lib/fluent/plugin/filter_kubernetes_metadata.rb
@@ -118,17 +118,21 @@ module Fluent::Plugin
       @stats.bump(:pod_cache_api_updates)
       log.trace("parsed metadata for #{namespace_name}/#{pod_name}: #{metadata}")
       @cache[metadata['pod_id']] = metadata
-    rescue StandardError => e
-      @stats.bump(:pod_cache_api_nil_error)
-      log.debug "Exception '#{e}' encountered fetching pod metadata from Kubernetes API #{@apiVersion} endpoint #{@kubernetes_url}"
-      if e.message == "Unauthorized"
-        @client = nil
-        # recreate client to refresh token
-        log.info("Re-creating Kubernetes API Client to refresh auth bearer token.")
-        create_client()
+      rescue KubeException => e
+        if e.error_code == 401
+          # recreate client to refresh token
+          log.info("Encountered '401 Unauthorized' exception, recreating client to refresh token")
+          create_client()
+        else
+          log.error "Exception '#{e}' encountered fetching pod metadata from Kubernetes API #{@apiVersion} endpoint #{@kubernetes_url}"
+          @stats.bump(:pod_cache_api_nil_error)
+        end
+        {}
+      rescue StandardError => e
+        @stats.bump(:pod_cache_api_nil_error)
+        log.error "Exception '#{e}' encountered fetching pod metadata from Kubernetes API #{@apiVersion} endpoint #{@kubernetes_url}"
+        {}
       end
-      {}
-    end
 
     def dump_stats
       @curr_time = Time.now
@@ -156,16 +160,20 @@ module Fluent::Plugin
       @stats.bump(:namespace_cache_api_updates)
       log.trace("parsed metadata for #{namespace_name}: #{metadata}")
       @namespace_cache[metadata['namespace_id']] = metadata
-    rescue StandardError => e
-      @stats.bump(:namespace_cache_api_nil_error)
-      log.debug "Exception '#{e}' encountered fetching namespace metadata from Kubernetes API #{@apiVersion} endpoint #{@kubernetes_url}"
-      if e.message == "Unauthorized"
-        @client = nil
-        # recreate client to refresh token
-        log.info("Re-creating Kubernetes API Client to refresh auth bearer token.")
-        create_client()
-      end
-      {}
+      rescue KubeException => e
+        if e.error_code == 401
+          # recreate client to refresh token
+          log.info("Encountered '401 Unauthorized' exception, recreating client to refresh token")
+          create_client()
+        else
+          log.error "Exception '#{e}' encountered fetching namespace metadata from Kubernetes API #{@apiVersion} endpoint #{@kubernetes_url}"
+          @stats.bump(:namespace_cache_api_nil_error)
+        end
+        {}
+      rescue StandardError => e
+        @stats.bump(:namespace_cache_api_nil_error)
+        log.error "Exception '#{e}' encountered fetching namespace metadata from Kubernetes API #{@apiVersion} endpoint #{@kubernetes_url}"
+        {}
     end
 
     def initialize
@@ -311,6 +319,7 @@ module Fluent::Plugin
 
     def create_client()
       log.debug 'Creating K8S client'
+      @client = nil
       @client = Kubeclient::Client.new(
         @kubernetes_url,
         @apiVersion,
--- a/lib/fluent/plugin/kubernetes_metadata_watch_namespaces.rb
+++ b/lib/fluent/plugin/kubernetes_metadata_watch_namespaces.rb
@@ -46,14 +46,14 @@ module KubernetesMetadata
         @stats.bump(:namespace_watch_gone_errors)
         log.info('410 Gone encountered. Restarting namespace watch to reset resource versions.', e)
         namespace_watcher = nil
-      rescue StandardError => e
-        if e.message == "Unauthorized"
-          @client = nil
+      rescue KubeException => e
+        if e.error_code == 401
           # recreate client to refresh token
-          log.info("Encountered 'Unauthorized' exception in watch, recreating client to refresh token")
+          log.info("Encountered '401 Unauthorized' exception in watch, recreating client to refresh token")
           create_client()
           namespace_watcher = nil
         else
+          # treat all other errors the same as StandardError, log, swallow and reset
           @stats.bump(:namespace_watch_failures)
           if Thread.current[:namespace_watch_retry_count] < @watch_retry_max_times
             # Instead of raising exceptions and crashing Fluentd, swallow
@@ -78,6 +78,31 @@ module KubernetesMetadata
             log.error(message, e)
             raise Fluent::UnrecoverableError, message
           end
+        end
+      rescue StandardError => e
+        @stats.bump(:namespace_watch_failures)
+        if Thread.current[:namespace_watch_retry_count] < @watch_retry_max_times
+          # Instead of raising exceptions and crashing Fluentd, swallow
+          # the exception and reset the watcher.
+          log.info(
+            'Exception encountered parsing namespace watch event. ' \
+            'The connection might have been closed. Sleeping for ' \
+            "#{Thread.current[:namespace_watch_retry_backoff_interval]} " \
+            'seconds and resetting the namespace watcher.', e
+          )
+          sleep(Thread.current[:namespace_watch_retry_backoff_interval])
+          Thread.current[:namespace_watch_retry_count] += 1
+          Thread.current[:namespace_watch_retry_backoff_interval] *= @watch_retry_exponential_backoff_base
+          namespace_watcher = nil
+        else
+          # Since retries failed for many times, log as errors instead
+          # of info and raise exceptions and trigger Fluentd to restart.
+          message =
+            'Exception encountered parsing namespace watch event. The ' \
+            'connection might have been closed. Retried ' \
+            "#{@watch_retry_max_times} times yet still failing. Restarting."
+          log.error(message, e)
+          raise Fluent::UnrecoverableError, message
         end
       end
     end
--- a/lib/fluent/plugin/kubernetes_metadata_watch_namespaces.rb
+++ b/lib/fluent/plugin/kubernetes_metadata_watch_namespaces.rb
@@ -47,29 +47,37 @@ module KubernetesMetadata
         log.info('410 Gone encountered. Restarting namespace watch to reset resource versions.', e)
         namespace_watcher = nil
       rescue StandardError => e
-        @stats.bump(:namespace_watch_failures)
-        if Thread.current[:namespace_watch_retry_count] < @watch_retry_max_times
-          # Instead of raising exceptions and crashing Fluentd, swallow
-          # the exception and reset the watcher.
-          log.info(
-            'Exception encountered parsing namespace watch event. ' \
-            'The connection might have been closed. Sleeping for ' \
-            "#{Thread.current[:namespace_watch_retry_backoff_interval]} " \
-            'seconds and resetting the namespace watcher.', e
-          )
-          sleep(Thread.current[:namespace_watch_retry_backoff_interval])
-          Thread.current[:namespace_watch_retry_count] += 1
-          Thread.current[:namespace_watch_retry_backoff_interval] *= @watch_retry_exponential_backoff_base
+        if e.message == "Unauthorized"
+          @client = nil
+          # recreate client to refresh token
+          log.info("Encountered 'Unauthorized' exception in watch, recreating client to refresh token")
+          create_client()
           namespace_watcher = nil
         else
-          # Since retries failed for many times, log as errors instead
-          # of info and raise exceptions and trigger Fluentd to restart.
-          message =
-            'Exception encountered parsing namespace watch event. The ' \
-            'connection might have been closed. Retried ' \
-            "#{@watch_retry_max_times} times yet still failing. Restarting."
-          log.error(message, e)
-          raise Fluent::UnrecoverableError, message
+          @stats.bump(:namespace_watch_failures)
+          if Thread.current[:namespace_watch_retry_count] < @watch_retry_max_times
+            # Instead of raising exceptions and crashing Fluentd, swallow
+            # the exception and reset the watcher.
+            log.info(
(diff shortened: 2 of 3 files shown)


PettitWesley:
@jcantrill Is this design for solving expired tokens acceptable for you? I've been testing it myself it works. I work for/with Amazon EKS and we have a lot of customers who need some sort of fix for expired tokens. Happy to change the code here to be whatever would suit your standards.

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -121,6 +121,12 @@ def fetch_pod_metadata(namespace_name, pod_name)
     rescue StandardError => e
       @stats.bump(:pod_cache_api_nil_error)
       log.debug "Exception '#{e}' encountered fetching pod metadata from Kubernetes API #{@apiVersion} endpoint #{@kubernetes_url}"
+      if e.message == "Unauthorized"
```
jcantrill, line 124: Is there a transport error of some kind that is throw where we can evaluate a response code (e.g. 401)?  This would seem to be more consistent

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -153,12 +159,20 @@ def fetch_namespace_metadata(namespace_name)
     rescue StandardError => e
       @stats.bump(:namespace_cache_api_nil_error)
       log.debug "Exception '#{e}' encountered fetching namespace metadata from Kubernetes API #{@apiVersion} endpoint #{@kubernetes_url}"
+      if e.message == "Unauthorized"
+        @client = nil
```
jcantrill, line 163: Maybe this should be in "create_client"

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -305,6 +309,17 @@ def configure(conf)
       end
     end
 
+    def create_client()
+      log.debug 'Creating K8S client'
```
jcantrill, line 313: I wonder if we need to make this function thread safe.?

lib/fluent/plugin/kubernetes_metadata_watch_namespaces.rb
```diff
@@ -47,29 +47,37 @@ def set_up_namespace_thread
...
-          log.info(
-            'Exception encountered parsing namespace watch event. ' \
-            'The connection might have been closed. Sleeping for ' \
-            "#{Thread.current[:namespace_watch_retry_backoff_interval]} " \
-            'seconds and resetting the namespace watcher.', e
-          )
-          sleep(Thread.current[:namespace_watch_retry_backoff_interval])
-          Thread.current[:namespace_watch_retry_count] += 1
-          Thread.current[:namespace_watch_retry_backoff_interval] *= @watch_retry_exponential_backoff_base
+        if e.message == "Unauthorized"
```
jcantrill, line 50: same question here regarding evaluating against a known code instead of a string

lib/fluent/plugin/filter_kubernetes_metadata.rb
PettitWesley: Let me explain how I chose this option. But first disclaimer- so to be honest I am not a ruby dev and I sort I actually mostly work on Fluent Bit but I was asked to work on it since its important to our customers but I mostly don't know what I'm doing here.

I had the same thought as you though, what's the most canonical way to match this specific error. 

So I ran this code and recorded what it outputted: https://github.com/PettitWesley/fluent-plugin-kubernetes_metadata_filter/blob/attempt_2/lib/fluent/plugin/kubernetes_metadata_watch_pods.rb#L53

And the result is this screenshot. Basically, in the string representation of the exception, which gets printed when you print it, there is an http code in the string (but not in the message or full_message fields, which is interesting). It felt wrong to match on that full string and the "Unauthorized" was easy to match on so I picked it. it didn't seem like there was an actual field on the exception object that would give me the code but may be I just didn't know how to find it. 

Screen Shot 2022-06-06 at 6 42 00 PM

lib/fluent/plugin/filter_kubernetes_metadata.rb
PettitWesley: (Unnecessary character building anecdote warning)

So fun fact: Once upon a time, when I was a new developer on the open source scene at my first conference, I got silly stuff from people on twitter because I made a big deal about how Fluent Bit is faster than Fluentd largely because of the language choice and ruby is very slow. 

And it has this thing called a Global Interpreter Lock, which means the multiple threads are never really every actually executing in parallel, its concurrency, not true parallelization, and it just means that when one thread is waiting on the network then another can do stuff, so its not possible for the watcher threads to try to refresh the client at the same time as the main thread.

PettitWesley: Hmm actually thinking about it some more... may be this isn't completely safe...

Can you help me understand how the watch functionality works- in watch mode, will the client calls in the main file (filter_kubernetes.rb) ever be made? 

I just had the thought that may be if one watch thread or the main thread refreshes the client, that will somehow break the watch on the other watch thread. But if that happens, the other thread would have run into an unauthorized exception anyway... so I think its fine actually. 

I think I just need to do some more testing of this on a real cluster with lots of updates happening over a decent period of time and then I can be confident its safe.

Can you think of any scenarios that this code change will likely break?

lib/fluent/plugin/filter_kubernetes_metadata.rb
jcantrill: @PettitWesley have you tried Rescuing on the KubeClient::HttpError ? https://github.com/ManageIQ/kubeclient/blob/master/lib/kubeclient/http_error.rb#L6

lib/fluent/plugin/filter_kubernetes_metadata.rb
jcantrill: > I just had the thought that may be if one watch thread or the main thread refreshes the client, that will somehow break the watch on the other watch thread. But if that happens, the other thread would have run into an unauthorized exception anyway... so I think its fine actually.

This was my original thought too.  The watch will more or less fail at the same time so you have a momentary blip in failure that will should converge back to a valid client capable of making calls.

lib/fluent/plugin/filter_kubernetes_metadata.rb
PettitWesley: My understanding was that the maintainers of that repo have not given any timeline for a release. Hence, its not something we can use to get a fix out to users ASAP- which is what AWS kubernetes users have requested and thus I am working on this. 

If there is any way that you can think of that we can reach out to them to speed up that release, or any way that this project can consume that code on master... its definitely preferable to my change.

larivierec: I don't see a problem with the PR at all. imo, I would probably add this fix using the error_code rather than message. it's never a bad thing to rely solely on underlying libraries.

PettitWesley: Cool!

> it's never a bad thing to rely solely on underlying libraries.

Wait I think you meant- it is a bad thing to rely solely on the underlying library, we should have protections for basic things like unauthorized exceptions in this code base as well? That's what I was thinking too after I gave it some more thought... I also just updated this PR with the `error_code` change and I'm testing it now..

PettitWesley:
I created this little test script and I am running it on a cluster where the token expiration is set to 1 hour. This should prove if it works and is safe or not:

```
while true
do
  kubectl apply -f app.yaml
  sleep 444
  POD=$(kubectl get pods -o name)
  kubectl label --overwrite  $POD status=hi
  sleep 444
  kubectl label --overwrite  $POD status=testing
  sleep 444
  kubectl label --overwrite  $POD status=hopefullyitworks
  sleep 444
  kubectl apply -f app2.yaml
  kubectl create namespace test
  sleep 444
  kubectl delete -f app.yaml
  POD=$(kubectl get pods -o name)
  kubectl label --overwrite  $POD status=hi
  sleep 444
  kubectl label --overwrite  $POD status=testing
  sleep 444
  kubectl label --overwrite  $POD status=hopefullyitworks
  sleep 444
  kubectl delete -f app2.yaml
  kubectl delete namespace test
  sleep 444
done
```

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -118,11 +118,20 @@ def fetch_pod_metadata(namespace_name, pod_name)
...
-    end
+      rescue KubeException => e
+        if e.error_code == 401
+          # recreate client to refresh token
+          log.info("Encountered '401 Unauthorized' exception, recreating client to refresh token")
+          create_client()
+        else
+          log.error "Exception '#{e}' encountered fetching pod metadata from Kubernetes API #{@apiVersion} endpoint #{@kubernetes_url}"
+          @stats.bump(:pod_cache_api_nil_error)
+        end
```
jcantrill, line 129: This looks like it needs to return `{}` after this end as callers are expecting a map.  I'm not certain otherwise what is the result upon error

lib/fluent/plugin/filter_kubernetes_metadata.rb
```diff
@@ -150,15 +159,26 @@ def fetch_namespace_metadata(namespace_name)
...
-      {}
+      rescue KubeException => e
+        if e.error_code == 401
+          # recreate client to refresh token
+          log.info("Encountered '401 Unauthorized' exception, recreating client to refresh token")
+          create_client()
+        else
+          log.error "Exception '#{e}' encountered fetching namespace metadata from Kubernetes API #{@apiVersion} endpoint #{@kubernetes_url}"
+          @stats.bump(:namespace_cache_api_nil_error)
+        end
```
jcantrill, line 170: same here with the return

lib/fluent/plugin/filter_kubernetes_metadata.rb
PettitWesley: Ah I see... I was wondering what the `{}` in the other rescue was for... will fix..

lib/fluent/plugin/filter_kubernetes_metadata.rb
PettitWesley: Fixed

PettitWesley:
@jcantrill I actually didn't think this was ready yet. I am still performing the final long running tests. 

Here is the output after running for almost 100 hours, on a cluster with the token expiration set to 1 hour, so the token would need to be refreshed about 100 times in this run- there are no errors or failures that I can see, and below is the current log output. This was with my script that continually restarts the same 2 pods and creates and deletes a namespace and changes labels on the pods. 

```
2022-06-26 00:55:17 +0000 [info]: #0 stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292
2022-06-26 00:55:17.846993520 +0000 fluent.info: {"message":"stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292"}
2022-06-26 00:55:51.846743737 +0000 fluent.info: {"message":"stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292"}
2022-06-26 00:55:51 +0000 [info]: #0 stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292
2022-06-26 00:56:27 +0000 [info]: #0 stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292
2022-06-26 00:56:27.847043383 +0000 fluent.info: {"message":"stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292"}
2022-06-26 00:57:03 +0000 [info]: #0 stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292
2022-06-26 00:57:03.848472347 +0000 fluent.info: {"message":"stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292"}
2022-06-26 00:57:37 +0000 [info]: #0 stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292
2022-06-26 00:57:37.846396057 +0000 fluent.info: {"message":"stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292"}
2022-06-26 00:58:13 +0000 [info]: #0 stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292
2022-06-26 00:58:13.850007351 +0000 fluent.info: {"message":"stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292"}
2022-06-26 00:58:49.849233695 +0000 fluent.info: {"message":"stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292"}
2022-06-26 00:58:49 +0000 [info]: #0 stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292
2022-06-26 00:59:23.850891448 +0000 fluent.info: {"message":"stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292"}
2022-06-26 00:59:23 +0000 [info]: #0 stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292
2022-06-26 00:59:59 +0000 [info]: #0 stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292
2022-06-26 00:59:59.847002656 +0000 fluent.info: {"message":"stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292"}
2022-06-26 01:00:35 +0000 [info]: #0 stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292
2022-06-26 01:00:35.847178135 +0000 fluent.info: {"message":"stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292"}
2022-06-26 01:01:09 +0000 [info]: #0 stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292
2022-06-26 01:01:09.847133076 +0000 fluent.info: {"message":"stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292"}
2022-06-26 01:01:45 +0000 [info]: #0 stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292
2022-06-26 01:01:45.847048708 +0000 fluent.info: {"message":"stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292"}
2022-06-26 01:02:21 +0000 [info]: #0 stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292
2022-06-26 01:02:21.845743474 +0000 fluent.info: {"message":"stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292"}
2022-06-26 01:02:55.847117226 +0000 fluent.info: {"message":"stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292"}
2022-06-26 01:02:55 +0000 [info]: #0 stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292
2022-06-26 01:03:31.846258958 +0000 fluent.info: {"message":"stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292"}
2022-06-26 01:03:31 +0000 [info]: #0 stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292
2022-06-26 01:04:07.851858517 +0000 fluent.info: {"message":"stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292"}
2022-06-26 01:04:07 +0000 [info]: #0 stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292
2022-06-26 01:04:41.846592617 +0000 fluent.info: {"message":"stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292"}
2022-06-26 01:04:41 +0000 [info]: #0 stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292
2022-06-26 01:05:18.847502683 +0000 fluent.info: {"message":"stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292"}
2022-06-26 01:05:18 +0000 [info]: #0 stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292
2022-06-26 01:05:54 +0000 [info]: #0 stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292
2022-06-26 01:05:54.854894194 +0000 fluent.info: {"message":"stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292"}
2022-06-26 01:06:28 +0000 [info]: #0 stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292
2022-06-26 01:06:28.848381563 +0000 fluent.info: {"message":"stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292"}
2022-06-26 01:07:04 +0000 [info]: #0 stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292
2022-06-26 01:07:04.856832648 +0000 fluent.info: {"message":"stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292"}
2022-06-26 01:07:40 +0000 [info]: #0 stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292
2022-06-26 01:07:40.846404396 +0000 fluent.info: {"message":"stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292"}
2022-06-26 01:08:14.846693561 +0000 fluent.info: {"message":"stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292"}
2022-06-26 01:08:14 +0000 [info]: #0 stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292
2022-06-26 01:08:50 +0000 [info]: #0 stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292
2022-06-26 01:08:50.846042861 +0000 fluent.info: {"message":"stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292"}
2022-06-26 01:09:26 +0000 [info]: #0 stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292
2022-06-26 01:09:26.847326543 +0000 fluent.info: {"message":"stats - namespace_cache_size: 3, pod_cache_size: 3, namespace_cache_api_updates: 292, pod_cache_api_updates: 292, id_cache_miss: 292"}
```

larivierec:
> This was with my script that continually restarts the same 2 pods and creates and deletes a namespace and changes labels on the pods. 

I'd like a small clarification, do you restart your pods? 
I don't believe this would be the right way, considering a restarted pod will always reload the token from the new file.

Would it be possible to test without restarting the pods?

PettitWesley:
@larivierec 

> I'd like a small clarification, do you restart your pods?


My script restarts the app pods, not the Fluentd pod. This tests fluentd picking up new pods which requires a new request to the API  server. See in my script the `kubectl delete -f app.yaml`  and `kubectl delete -f app2.yaml`- those are my app pods. My Fluentd pod is `fluentd.yaml` and it was running continuously without restarts for the 100 hours. I checked.

larivierec:
Alright thanks for the clarification, just wanted to make sure! 👍

PettitWesley:
**OK actually I might have panicked too much, I think it fully works I just need to redo the testing to feel safe.**

PettitWesley:
@larivierec Yea so this is why my testing was invalid. The code I wrote is **reactive** not **proactive**, it refreshes when the token expires, not when its stale. Since the 1 hour token refresh on my cluster was removed due to a communication mistake, my testing isn't valid since the token never expired in my latest tests (I did do a test earlier with the 1 hour expiration and it worked but then I made a few changes and in the earlier test I didn't check changing labels and all the things my script docs... hence I think it works but I can't be certain yet...)

PettitWesley:
We have fixed our cluster to expire tokens in 1 hour. So we will know soon if this works. I'm not too worried...

PettitWesley:
We have now validated that the `watch true` behavior works... so it should be good... to be safe I will let this test run a bit longer and then put in a test for `watch false` over night.

PettitWesley:
And we have now validated that `watch false` fully works! Sorry for the panic! But thorough pre-release testing is always important!

harshit-splunk:
@PettitWesley from what I have understood, the token will be updated when the client receives an unauthorised error. So it is expected that we will see `seconds after warning threshold` in API server audit logs once every hour.

smrutimandal:
@PettitWesley Thanks for this. Just to clarify, does it also mean that for EKS 1.21, the token won't be refreshed for 90 days as the kube-apiserver doesn't throw a 401 for the stale tokens for 90 days?

I am running the following:

```
eks: 1.21
fluentd image: v1.14.6-debian-1.1
fluent-plugin-kubernetes_metadata_filter gem: 2.11.1
```

PettitWesley:
@smrutimandal Correct.

Ga13Ou:
> We have fixed our cluster to expire tokens in 1 hour. So we will know soon if this works. I'm not too worried...

@PettitWesley Can you tell how did you manage to configure your EKS cluster to expire token in 1 hour? it seems like such configuration can only be done at apiserver level, which is not customizable when using EK

PettitWesley:
Sorry, I no longer remember.

PettitWesley:
There was some API command that allowed us to set that. Search docs and github.

PR #349, opened 2022-07-20
fix testapiadapter

jcantrill:
This PR:
* fixes the TestApiAdapter which relied upon  String#underscore


jcantrill:
Confirmed with functional test from OpenShift Log test

PR #350, opened 2022-07-29
Reduce kubeclient timeouts to reasonable values

orgads:
The API should be very responsive. If it hangs, it is usually because
something is wrong, and waiting longer won't help.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,8 @@ when true (default: `true`)
 * `skip_namespace_metadata` - Skip the namespace_id field from the metadata. The fetch_namespace_metadata function will be skipped. The plugin will be faster and cpu consumption will be less.
 * `stats_interval` - The interval to display cache stats (default: 30s).  Set to 0 to disable stats collection and logging
 * `watch_retry_interval` - The time interval in seconds for retry backoffs when watch connections fail. (default: `10`)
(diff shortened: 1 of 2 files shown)


orgads:
We had a pod that got a reply for `get_pod`, but timed out (60s) on `get_namespace`. Ideally, I'd prefer to retry 2-3 times on such cases. What do you say?

What's the preferred way to retry?

jcantrill:
> We had a pod that got a reply for `get_pod`, but timed out (60s) on `get_namespace`. Ideally, I'd prefer to retry 2-3 times on such cases. What do you say?

This may be true but what about cases with networks which have high latency?  You may begin to have throughput issues but retrying won't help in these cases. 

> 
> What's the preferred way to retry?

I believe the better way to resolve this issue is to introduce a configuration parameter that allows user's to tweak this setting instead of assuming the defaults provided here fit everyones use case.  These may be fine for defaults but we should also be able to override them.

orgads:
Done (no retry yet).

jcantrill:
Looks good.  Please also update the README

orgads:
> Looks good. Please also update the README

Thank you, done.

jcantrill:
https://rubygems.org/gems/fluent-plugin-kubernetes_metadata_filter/versions/3.1.0

orgads:
Thank you

PR #352, opened 2022-09-07
pod watcher must reinitialize on 401

steved:
The unauthorized error handling added in https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/pull/337 correctly handles re-initializing the namespace watcher, but a dangling reference to it was left in the pod watcher causing it to infinitely loop with an expired client.


jcantrill:
https://rubygems.org/gems/fluent-plugin-kubernetes_metadata_filter/versions/3.1.1

PR #354, opened 2022-09-30
LOG-3046: clone meta so the ref can't be modified for subsequent records

jcantrill:
This PR:
* dups  the meta for cases where it gets modified in processing and we dont want  it to affect future filtering

ref: https://issues.redhat.com/browse/LOG-3046

cc @cahartma


cahartma (commented):
/lgtm

PR #357, opened 2022-11-23
Reduce log level for missing pod error

smlx:
It is not uncommon for a pod to be gone by the time logs are processed, so reduce the log level appropriately. This avoids emitting spurious error messages for harmless events.

The log level for this kind of API error was `debug` in v2, but was increased to `error` in [this PR](https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/pull/337/files#diff-0b4b9044f59b6776e7937e8f714cb59db8858c728efb76dd18456f026f423f19R127) which made it into v3. So this change just reverts the logging behaviour back to what it was in v2.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -79,7 +79,7 @@ GEM
     parallel (1.22.1)
     parser (3.1.2.0)
       ast (~> 2.4.1)
-    power_assert (2.0.1)
+    power_assert (2.0.2)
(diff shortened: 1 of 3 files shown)


jcantrill:
https://rubygems.org/gems/fluent-plugin-kubernetes_metadata_filter/versions/3.1.3

PR #371, opened 2023-07-21
Improve pod and namespace label management

forselli-stratio:
Hi!

Currently, the plugin allows to obtain labels from the pods and the namespaces or exclusively the ones from the pods (if I use the parameter skip_namespace_metadata: true and skip_labels: false).
In multi-tenant scenarios, we may need namespace labels to be able to assign each event to its tenant and, with the current configuration options, the only way to get these labels is by getting the ones from the pods too. In an average cluster, i think it is more likely that the labels of the pods may not be standardized and therefore we have too many. 

The idea of ​​this PR is to allow to choose between the labels from the pod or from the namespace, and that you don't always have to get both.

Please let me know what do you think about this feature or if i'm missing something. If you consider this is a good idea, I'll continue working on the PR to implement tests and any other suggestion you have (first time with ruby, so keep that in mind ;) )

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,8 @@ when true (default: `true`)
   to trust the intermediate CA certs we do have, set this to `true` - this corresponds to
   the `openssl s_client -partial_chain` flag and `X509_V_FLAG_PARTIAL_CHAIN` (default: `false`)
 * `skip_labels` - Skip all label fields from the metadata.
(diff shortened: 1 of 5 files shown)


jcantrill:
If you find this feature useful please continue to add tests, etc.  Assuming there are no regressions I have no reason to not accept it.

PR #375, opened 2023-09-25
feat: add kubernetes controller metadata

aauren:
This PR adds controller metadata to the list of available data that could be retrieved by the metadata filter. It is defaulted to false so that the default functionality of the plugin is not changed and people that do not want this information are not impacted by it.

When set to true, it produces an integer keyed map of the list of controllers that own / manage the pod. This information is available from the API server via the following:

```sh
kubectl get pod <foo> -o 'jsonpath={.metadata.ownerReferences}'
```

It will add something like the following to each record (flattened for readability):

```
kubernetes.controller.0.kind = ReplicaSet
kubernetes.controller.0.name = pong-672fb1239d
```

This information is useful for those wanting to get more direct references to the higher level objects in k8s like ReplicaSets, DaemonSets, Jobs, etc. from their logs without having to make / enforce strict labeling policies.

My Ruby is a bit rusty, so feel free to let me and I'll make any changes required. I have tested this and am running the plugin with this patch in a medium sized development cluster and it so far works fine.

### DIFF
--- a/lib/fluent/plugin/kubernetes_metadata_common.rb
+++ b/lib/fluent/plugin/kubernetes_metadata_common.rb
@@ -82,14 +82,27 @@ module KubernetesMetadata
         log.warn("parsing container meta information failed for: #{pod_object[:metadata][:namespace]}/#{pod_object[:metadata][:name]}: #{e}")
       end
 
+      ownerrefs_meta = []
+      begin
+        pod_object[:metadata][:ownerReferences].each do |owner_reference|
+          ownerrefs_meta.append({
+            'kind' => owner_reference[:kind],
+            'name' => owner_reference[:name]
+          })
+        end
+      rescue StandardError => e
+        log.warn("parsing ownerrefs meta information failed for: #{pod_object[:metadata][:namespace]}/#{pod_object[:metadata][:name]}: #{e}")
+      end if @include_ownerrefs_metadata && pod_object[:metadata][:ownerReferences]
+
       kubernetes_metadata = {
         'namespace_name' => pod_object[:metadata][:namespace],
         'pod_id' => pod_object[:metadata][:uid],
         'pod_name' => pod_object[:metadata][:name],
         'pod_ip' => pod_object[:status][:podIP],
         'containers' => syms_to_strs(container_meta),
-        'host' => pod_object[:spec][:nodeName]
-      }
+        'host' => pod_object[:spec][:nodeName],
+        'ownerrefs' => (ownerrefs_meta if @include_ownerrefs_metadata)
+      }.compact
       kubernetes_metadata['annotations'] = annotations unless annotations.empty?
       kubernetes_metadata['labels'] = labels unless labels.empty?
       kubernetes_metadata['master_url'] = @kubernetes_url unless @skip_master_url
--- a/README.md
+++ b/README.md
@@ -64,6 +64,7 @@ when true (default: `true`)
 * `watch_retry_interval` - The time interval in seconds for retry backoffs when watch connections fail. (default: `10`)
 * `open_timeout` - The time in seconds to wait for a connection to kubernetes service. (default: `3`)
 * `read_timeout` - The time in seconds to wait for a read from kubernetes service. (default: `10`)
+* `include_ownerrefs_metadata` - If set to true, it will include metadata (`kind` & `name`) in `kubernetes.ownerrefs` about the controller that owns the pod. (default: `false`)
 
 
 Reading from a JSON formatted log files with `in_tail` and wildcard filenames while respecting the CRI-o log format with the same config you need the fluent-plugin "multi-format-parser":
--- a/lib/fluent/plugin/filter_kubernetes_metadata.rb
+++ b/lib/fluent/plugin/filter_kubernetes_metadata.rb
@@ -82,6 +82,7 @@ module Fluent::Plugin
     config_param :skip_container_metadata, :bool, default: false
     config_param :skip_master_url, :bool, default: false
     config_param :skip_namespace_metadata, :bool, default: false
+    config_param :include_ownerrefs_metadata, :bool, default: false
 
     # A classname in the form of Test::APIAdapter which will try
     # to be resolved from a relative named file 'test_api_adapter'
--- a/test/cassettes/kubernetes_get_pod_with_ownerrefs.yml
+++ b/test/cassettes/kubernetes_get_pod_with_ownerrefs.yml
@@ -1,0 +1,156 @@
+#
+# Fluentd Kubernetes Metadata Filter Plugin - Enrich Fluentd events with
+# Kubernetes metadata
+#
+# Copyright 2015 Red Hat, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+---
+http_interactions:
+- request:
+    method: get
+    uri: https://localhost:8443/api/v1/namespaces/default/pods/fabric8-console-controller-98rqc
+    body:
+      encoding: US-ASCII
+      string: ''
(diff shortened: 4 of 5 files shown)


lib/fluent/plugin/kubernetes_metadata_common.rb
```diff
@@ -82,13 +82,28 @@ def parse_pod_metadata(pod_object)
...
+
       kubernetes_metadata = {
         'namespace_name' => pod_object[:metadata][:namespace],
         'pod_id' => pod_object[:metadata][:uid],
         'pod_name' => pod_object[:metadata][:name],
         'pod_ip' => pod_object[:status][:podIP],
         'containers' => syms_to_strs(container_meta),
-        'host' => pod_object[:spec][:nodeName]
+        'host' => pod_object[:spec][:nodeName],
+        'controller' => syms_to_strs(controller_meta)
```
jcantrill, line 106: might this better be called "ownerrefs" or something of the like?  It otherwise looks good to me and could use at least one test

aauren: Thanks for the quick review!

Do you want me to change the option name as well? from `include_controller_metadata` to `include_owner_metadata` or some such?

jcantrill: Please update the config to make it clear what it is enabling

aauren: @jcantrill alright, I think I have implemented all of your recommendations.

* I changed everything to be consistently `ownerrefs` rather than `controller`.
* I ensured that the existing tests were passing
* I added an additional test for this new attribute

Let me know if you want any further changes.

lib/fluent/plugin/kubernetes_metadata_common.rb
```diff
@@ -82,14 +82,29 @@ def parse_pod_metadata(pod_object)
         log.warn("parsing container meta information failed for: #{pod_object[:metadata][:namespace]}/#{pod_object[:metadata][:name]}: #{e}")
       end
 
+      ownerrefs_meta = {}
+      begin
+        key = 0
+        pod_object[:metadata][:ownerReferences]. each do |owner_reference|
+          ownerrefs_meta[key.to_s] = {
```
jcantrill, line 89: Thinking about this more, why does this need to be a map where the key is a number?  Why can it not be implemented as an array? I don't see the need for map given its not a "named" thing

aauren: I originally started it as an array, but I didn't like the data representation on downstream systems. When it got ingested into our OpenSearch store, it ended up being decoded as a JSON blob rather than a field. e.g. I would end up with something like:

```
kubernetes.pod_name = foo-123151-adf12411
kubernetes.ownerrefs = {"kind": "ReplicaSet", "name": "foo-123151"}
```

I found this much harder to work with. Many of our users are used to maps, because every other field that comes through fluentd is a map. Additionally, the above is a lot harder to search on or parse than the other mapped fields that the metadata plugin produces. Finally, in this method, it gives me a consistent way to reference the primary owner via search fields e.g. `kubernetes.ownerrefs.0.kind` or some such.

This may be somewhat specific to our use case. I'm not sure how all of the other downstream log stores would handle the above, I only know about the ElasticSearch / Kibana family of tools. However, I would guess that without a consistent key field a lot of the other tools would struggle here as well. In general I would say that since all of the other fields that we produce are maps, it is better to go this way for consistency.

alanconway: @aauren - I don't understand what you mean by "decode as JSON blob", can you illustrate with some more complete examples?

If I understand your map looks like:
```
{
  "kubernetes": {
    "controller": {
      "0": { "kind":"foo", ...}
      "1": { "kind":"bar", ...}
   }
}
```
And the path expression to extract the kind field looks like:
```
kubernetes.controller.0.kind
```

As an array this looks like:
```
{
  "kubernetes": {
    "controllers": [
      { "kind":"foo", ...}
      { "kind":"bar", ...}
   ]
}
```
And the path expression to extract the kind field looks like:
```
kubernetes.controller[0].kind
```

I don't see how the map version is easier to use or understand. Unless the integer keys mean something, I don't think they should be required. If controllers have a unique string name or ID then that would be a good map key, but if you want a collection of objects with no meaningful keys then I think an array is the way to go.

I may be missing something, please clarify with examples.

aauren: It has to do with downstream parsers. Things like ElasticSearch don't know what to do with Array fields.

If you're only parsing it with other fluentd plugins, then yes, you're correct that the array is just as simple to index as a map.

However, if you're using a downstream search store, they're used to receiving data in maps rather than arrays. See https://www.elastic.co/guide/en/elasticsearch/reference/current/array.html specifically

```
Arrays of objects do not work as you would expect: you cannot query each object independently of the other objects in the array. If you need to be able to do this then you should use the [nested](https://www.elastic.co/guide/en/elasticsearch/reference/current/nested.html) data type instead of the [object](https://www.elastic.co/guide/en/elasticsearch/reference/current/object.html) data type.
```

Additionally, every other field that fluent-plugin-kubernetes_metadata_filter produces is a map, so I think that this fits the principle of least surprise when people are enabling the feature.

aauren: Note the above document mentions "nested" fields which is a way to change the mapping in ES to handle arrays. However, this requires that the user manually apply a mapping ahead of time before enabling the feature, and it is also ES specific, finally it essentially does the same thing in the background of making it a map internally.

I would argue that just emitting it as a number indexed map is simpler all the way around unless there is some implicit cost that we're worried about with using a hash vs an array in Ruby.

Additionally, the mapping option is ES specific. Emitting it as a map in the first place is more likely to work with other logging systems out of the box similar to other fields that are already emitted by the plugin.

xperimental: If the int-map is a specific workaround for ES, would it make sense to use the array notation in the plugin here where it's not tied to a particular output and do the "ES specific" processing later in a filter before shipping the entries to the output?

jcantrill: My ask was specifically geared towards what is the better API and not output specific.  The fact that the other fields are received as a map seems immaterial from an API perspective; they are truly mapped entities and the representation should, IMO, represent how they are received.  I would prefer to not design the API with Elasticsearch in mind.  There are tools (e.g. jq) that are capable of filtering a nested array.

> I would argue that just emitting it as a number indexed map is simpler all the way around unless there is some implicit cost that we're worried about with using a hash vs an array in Ruby.

This is an API question and nothing to do with the cost in Ruby.  Ruby is quite good at spinning through enumerations and I suspect whether it is `map.each do |k,v|` or `array.each_with_index do |v,i|` they are roughly the same

aauren: My thought is that there are likely other log stores that behave similarly although I don't have much experience with other stores. I can definitely say that reducing it to an array would be much harder on our use case, but I'm unable to say that it would be that way for all users. I usually tend towards being more pragmatic with this type of thing.

Everyone else in this thread seems to think Arrays would be better, so I'm willing to make that change if that's what would make this work mergable. Data in a bad format for our use-case would still be better than no data at all.

alanconway:
@aauren a couple thoughts that may or may not help.  
The ES docs say "you cannot query each object independently of the other objects in the array."

Arrays are used in 2 cases:
1. You have a list with a known order, you need to preserve that order.
2. You don't care about order, but there is no unique key you can use to identify objects.

I think you are in case 2, and regardless of ES restrictions that means that: _there is no meaningful way to query objects independently anyway_. You cannot know the numbering in advance, so you can't formulate a query to get a desired value based on its number. You can only grab the whole array and then iterate to find the thing you want. It is true that once you have iterated over the array,  you could "remember" indexes and use them to retrieve specific objects; but why would you do that when you have already downloaded the whole array?

So is it really a problem? 

Cheers,
Alan.

lib/fluent/plugin/kubernetes_metadata_common.rb
aauren: I have reverted the data type to be an array and updated the tests to resolve the concerns that have been mentioned here.

Please let me know if there are any other required changes to make this mergeable.

jcantrill:
https://rubygems.org/gems/fluent-plugin-kubernetes_metadata_filter/versions/3.4.0

aauren:
@jcantrill thanks for the prompt PR feedback and cutting the new release!

PR #405, opened 2025-08-20
Update metadata because of transfered repository

kenhys:
https://github.com/fluent-plugins-nursery/contact/issues/33


kenhys:
Thanks.

PR #407, opened 2025-08-21
v3.8.0

kenhys:
As Fluentd v1.19.0 was released, let's ship v3.8.0.


kenhys:
--minor seems too aggressive.

kenhys:
Use bundle update --patch to respect previous gem updating policy. (I guess)

Watson1978 (approved):
👍🏻

kenhys:
Thank you for reviewing

PR #425, opened 2026-06-11
Remove unused yajl dependency

Watson1978:
Since the yajl-ruby gem depends on a deprecated Ruby C API, it cannot be installed with Ruby 4.1.


kenhys (approved):

kenhys:
Good catch.

